PR #7, opened 2022-11-03
Premises characteristics can be saved

bryangaledxw:
# Changes in this PR

* A user can enter characteristics for a premises
* A user can see characteristics for a premises

### Before
localhost_3000_properties_d84c6ad8-7334-4fcd-8191-ec6784af0844 (1)
localhost_3000_properties_new (2)

### After
localhost_3000_properties_d84c6ad8-7334-4fcd-8191-ec6784af0844
localhost_3000_properties_new (1)

### DIFF
--- a/e2e/tests/stepDefinitions/manage.ts
+++ b/e2e/tests/stepDefinitions/manage.ts
@@ -26,13 +26,7 @@ Given("I'm creating a premises", () => {
 Given('I create a premises with all necessary details', () => {
   const page = PremisesNewPage.verifyOnPage(PremisesNewPage)
 
-  cy.get('select[name="localAuthorityAreaId"').within(() => {
-    cy.get('option')
-      .contains('North Lanarkshire')
-      .then(element => {
-        cy.wrap(element.attr('value')).as('localAuthorityAreaId')
-      })
-  })
+  page.getLocalAuthorityAreaIdByLabel('North Lanarkshire', 'localAuthorityAreaId')
 
   cy.get('@localAuthorityAreaId').then(localAuthorityAreaId => {
     const newPremises = newPremisesFactory.build({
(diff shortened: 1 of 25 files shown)


bryangaledxw:
[Premises characteristics can be saved](https://trello.com/c/cOmYoJ8T/476-premises-characteristics-can-be-saved)

server/utils/characteristicUtils.ts
```diff
@@ -0,0 +1,11 @@
+import { Characteristic } from '../@types/shared'
+
+export const prepareCharacteristics = (
```
richpjames: 'prepare' is a bit vague as a name to me but I can't suggest anything better and the function body is very clear so I think it's fine for now.

e2e/tests/stepDefinitions/manage.ts
```diff
@@ -28,10 +29,17 @@ Given('I create a premises with all necessary details', () => {
 
   page.getLocalAuthorityAreaIdByLabel('North Lanarkshire', 'localAuthorityAreaId')
 
-  cy.get('@localAuthorityAreaId').then(localAuthorityAreaId => {
+  page.getCharacteristicIdByLabel('Park nearby', 'parkNearbyCharacteristicId')
+  page.getCharacteristicIdByLabel('Floor level access', 'floorLevelAccessCharacteristicId')
+
+  cy.then(function _() {
```
richpjames, line 35: Why the underscore? keeping the linter happy?

richpjames (dismissed):
LGTM. 
Really well structured and described PR - thanks Bryan!

e2e/tests/stepDefinitions/manage.ts
bryangaledxw: Basically 😭

Can't use arrow notation here, as then you lose access to the function's `this` object, which I'm using to get the aliased values in a tidier way than lots of nested `cy.get('@somealias').then()` calls. And there's the convention of using `_` as a prefix for function param names that are unused but needed to satisfy a signature, or for unused variables when destructuring a tuple

server/utils/characteristicUtils.ts
bryangaledxw: Yeah, I don't like it but best I could come up with was that or filterAndSortCharacteristics

bryangaledxw: Actually, typing that out, maybe that's fine, brb 😛

richpjames (approved):

PR #15, opened 2022-11-09
Remove getService()

bryangaledxw:
As Temporary Accommodation UI is now its own project, we can remove `getService`, and some no-longer used configuration options


pezholio (approved):
LGTM 👍

PR #19, opened 2022-11-11
A user can view a single bedspace

bryangaledxw:
# Changes in this PR

- A user can view a page showing the details of a single bedspace, rather than only seeing them in a list on the property page
- Editing a bedspace is now reached via this new page

### Before

localhost_3000_properties_0d0b2e17-c2ae-4ea1-b64d-c642e2c130a5 (1)

### After
localhost_3000_properties_0d0b2e17-c2ae-4ea1-b64d-c642e2c130a5

localhost_3000_properties_0d0b2e17-c2ae-4ea1-b64d-c642e2c130a5_bedspaces_8f0ee60d-482a-4086-a4d7-d85ce53e6b93

### DIFF
--- a/cypress_shared/pages/temporary-accommodation/manage/premisesShow.ts
+++ b/cypress_shared/pages/temporary-accommodation/manage/premisesShow.ts
@@ -80,12 +80,12 @@ export default class PremisesShowPage extends Page {
     })
   }
 
-  clickBedpaceEditLink(room: Room): void {
+  clickBedpaceViewLink(room: Room): void {
     cy.get('h4')
       .contains(room.name)
       .parents('[data-cy-bedspace]')
       .within(() => {
-        cy.get('a').contains('Edit').click()
+        cy.get('a').contains('View').click()
       })
   }
 }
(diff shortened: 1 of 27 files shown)


bryangaledxw:
[A user can view a single bedspace](https://trello.com/c/3chs6irD/519-a-user-can-view-a-single-bedspace)

cypress_shared/pages/temporary-accommodation/manage/premisesShow.ts
```diff
@@ -5,7 +5,7 @@ import paths from '../../../../server/paths/temporary-accommodation/manage'
 
 export default class PremisesShowPage extends Page {
   constructor(private readonly premises: Premises) {
-    super('Manage a property')
+    super('View a property')
```
richpjames, line 8: The commit message for this change is the wrong way round: 'Change "manage a property" to "view a property" is correct

server/services/premisesService.ts
```diff
@@ -80,6 +80,13 @@ export default class PremisesService {
       })
   }
 
+  async getPremises(token: string, id: string): Promise<Premises> {
+    const premisesClient = this.premisesClientFactory(token)
+    const premises = (await premisesClient.find(id)) as Premises
```
richpjames: Minor, but do we need to cast this? I would've thought the return type could be inferred from the `find` method

richpjames (approved):

server/services/premisesService.ts
bryangaledxw: Ah, yep, we don't, it's already coming from the client with the correct type. Thanks!

cypress_shared/pages/temporary-accommodation/manage/premisesShow.ts
bryangaledxw: 👍

PR #20, opened 2022-11-13
Added ADRs for data importing

Floppy:
# Context

Trello card: https://trello.com/c/nnbGhJ2A

# Changes in this PR

Two ADRs added which detail the design of the data conversion and import process.

### DIFF
--- a/doc/architecture/decisions/0006-manually-convert-data-for-import.md
+++ b/doc/architecture/decisions/0006-manually-convert-data-for-import.md
@@ -9,8 +9,8 @@ Accepted
 ## Context
 
 HPTs that are already running CAS3 services have a set of existing data about their properties, people,
-and so on, that they will need to add to the new CAS3 service. This existing data is stored in a set
-of spreadsheets.
+and so on, that they will need to add to the new CAS3 service. This existing data is stored in a set of
+spreadsheets.
 
 To avoid errors, and increase speed of transfer, we want to support conversion of the existing data,
 and avoid rekeying of that data into our new system.
@@ -39,7 +39,7 @@ different formats that will only be used once.
 CSV is chosen as the interim file format because it is simple to parse in code, and forces plain text, removing
 formatting.
 
-Because of the different data structures involved, there are likely to be separate intermediate formats for
+Because of the data structures involved, there are likely to be separate intermediate formats for
 property details, and for people and their placements into those properties.
 
 ## Consequences
@@ -58,3 +58,6 @@ behaviour. Any errors during conversion will thus be able to be corrected later 
 
 We do not anticipate loading past data in, only current state, but this should be checked with stakeholders
 and the data schema designed with the finding in mind.
+
+When the initial property data load is done, the CAS3 system will not yet be supporting tracking of bookings,
+so this will likely be done in a subsequent phase, either automatically or manually if necessary.
--- a/doc/architecture/decisions/0006-manually-convert-data-for-import.md
+++ b/doc/architecture/decisions/0006-manually-convert-data-for-import.md
@@ -1,0 +1,60 @@
+# 6. Manually Convert Data For Import
+
+Date: 2022-11-01
+
+## Status
+
+Accepted
+
+## Context
+
+HPTs that are already running CAS3 services have a set of existing data about their properties, people,
+and so on, that they will need to add to the new CAS3 service. This existing data is stored in a set
+of spreadsheets.
+
+To avoid errors, and increase speed of transfer, we want to support conversion of the existing data,
+and avoid rekeying of that data into our new system.
+
+The existing spreadsheets are designed by each HPT to match their needs and design during the prototype
+phase of the service rollout; they are not consistent across areas. We have used those spreadsheets to
+inform our database design, but there is not necessarily a direct field mapping for every data item due
+to this variation. For instance, what might be a boolean field for "wheelchair access" in our database might
+be a YES/NO column in one area, or a freeform text note in another.
+
+The challenge is therefore to transform the existing data into a common format based on our database schema.
+
+## Decision
+
+Due to the complexity and variety of the input data, we do not propose to create an automatic conversion process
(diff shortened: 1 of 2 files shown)


doc/architecture/decisions/0006-manually-convert-data-for-import.md
```diff
@@ -0,0 +1,60 @@
...
+
+Date: 2022-11-01
+
+## Status
+
+Accepted
+
+## Context
+
+HPTs that are already running CAS3 services have a set of existing data about their properties, people,
```
tahb, line 11: Perhaps we should be explicit that we aren't including/thinking about booking data? This would need to be manually added after this property data is imported.

doc/architecture/decisions/0007-load-data-in-via-api.md
```diff
@@ -0,0 +1,53 @@
...
+## Status
+
+Accepted
+
+## Context
+
+We need to import existing data into the CAS3 system in order for HPTs that are currently providing the service
+to be able to carry on doing so with minimal interruption.
+
+Data needs to be loaded into our production environment quickly and automatically. This data is stored in
```
tahb: I imagine we will want to add data into the 'pre-production' environment as well so there is an element of repeatability here.

doc/architecture/decisions/0007-load-data-in-via-api.md
```diff
@@ -0,0 +1,53 @@
...
+
+Date: 2022-11-01
+
+## Status
+
+Accepted
+
+## Context
+
+We need to import existing data into the CAS3 system in order for HPTs that are currently providing the service
```
tahb, line 11: We have approximate maxiumum of 2500 bedspaces across 12 regions. It may be useful context to share that we expect to onboard one region at a time with each region having 100-200 bedspaces.

doc/architecture/decisions/0007-load-data-in-via-api.md
```diff
@@ -0,0 +1,53 @@
...
+* The domain API can enforce all its business logic on the inserted data, ensuring security and consistency.
+* We can prove that the API works correctly by exercising it with the imported data.
+
+The import process will be tested fully, and the data verified, in a pre-production environment before running the import on production.
+
+## Consequences
+
+During the import, we will want to disable any notifications that are sent by the domain API backend, so that users are not confused about being notified things they already know about.
+
+The import script should be written so that it is idempotent; errors will inevitably creep in, and if the import process can simply be rerun without duplicating already-entered data, this process will be much simpler.
```
tahb, line 44: I wonder how we could recover if we made a mistake and needed to start again for all or a single region?

doc/architecture/decisions/0007-load-data-in-via-api.md
```diff
@@ -0,0 +1,53 @@
...
+
+It may be simpler to build the import process into the UI itself, rather than as a separate script, seeing as
+the UI already connects to the API and performs many of the same operations. That is left as an implementation
+detail for the development team to decide.
```
tahb, line 48: This has got me thinking.. I wonder if we could actually build a new temporary frontend (maybe in Ruby so I can help) that takes in a CSV into memory for processing. We might be able to control access via hmpps-auth so that only devs can sign into it and then we can remove the whole service when it's done with a nice degree of separation. 

It's perhaps a reach but perhaps this could also be an interface to managing dynamic lists such as `characteristics`, `move_on_categories ` etc.

tahb: I've shared these thoughts with CAS1 https://trello.com/c/SxGVR04d/830-seeding-bootstrapping-data-work#comment-637248da7373df05d98b83f0

doc/architecture/decisions/0007-load-data-in-via-api.md
```diff
@@ -0,0 +1,53 @@
...
+An import script will need to be written to parse the intermediate format, and perform this insertion.
+
+We have considered two options for the import script:
+
+1. Data files are copied to a node inside our production environment, and an import script is run within the privileged environment to load the data, either via Kotlin backend code or direct SQL statements.
+2. Data files are copied to a secure machine outside the production environment, and an import script uses the domain API (which is also used by the UI for all its tasks) to insert data.
+
+## Decision
+
+We will insert data via the domain API, with an import script running on a secure but non-production machine.
```
tahb, line 30: Have we spoken to others in confirming this approach is acceptable? It might be useful context to understand what endorsement there is.

tahb (approved):
Great write up James. I've left some minor comments including a couple of questions where it would be great to add a bit more if we have it?

doc/architecture/decisions/0006-manually-convert-data-for-import.md
Floppy: Added a comment in consequences about that. :+1:

doc/architecture/decisions/0007-load-data-in-via-api.md
Floppy: added

doc/architecture/decisions/0007-load-data-in-via-api.md
Floppy: added

doc/architecture/decisions/0007-load-data-in-via-api.md
Floppy: I can't think of a sensible approach other than manually clearing it; finding all data for a region would be easy enough if it was just one area. Not sure it's worth automating, and one disadvantage of the API method is that we can't wrap it in a transaction.

doc/architecture/decisions/0007-load-data-in-via-api.md
Floppy: Added a note about this in a completed item in the actions section.

Floppy:
@tahb this is good to go now, I've addressed your comments. I can't merge it in myself, so if you're happy, go ahead.

PR #23, opened 2022-11-16
Delete Approved Premises Controllers

bryangaledxw:
In this PR we delete used Approved Premises controllers, which are no longer useful to have in the Temporary Accommodation UI project. This is a quick incremental step in removing legacy AP code from the project. This does leave many unused utility and service functions. It also does not un-namespace TA controllers/routes/templates, for the sake of not disrupting in-progress work

This will though make it easier to remove unused functions in future. In particular, we will no longer need to keep them working against future shared type updates


pezholio (approved):
LGTM 👍

PR #56, opened 2022-12-07
Refactor reference data handling

bryangaledxw:
In this PR we refactor our reference data handling to be more in line with AP, and to be internally consistent with TA

- We synchronise our characteristics and local authorities JSON data with the API. This lets us simplify our E2E tests, as Cypress no longer needs to grab reference data entry IDs from the page in order to build a valid Premises/Room object
- We remove reference data type specific clients/services, and have other services return a bundle of relevant reference data


pezholio:
Got it now - Only it seems the integration tests are failing here

pezholio (dismissed):
LGTM - thanks for taking the time to split this up 👍

pezholio (approved):

PR #58, opened 2022-12-08
Remove the null allowlist for preprod

pezholio:
Even though we’re not mastering data in preprod yet, we are accessing preprod services, so should be using the master allowlist in the main `hmpps-temporary—accommodation-ui/values.yaml` file


tahb (approved):
We can sign in on preprod for TA (though the redirect endpoints are still wrong and are being updated). Let's see what happens now that's confirmed. Thanks for doing this for us Stu ❤️

PR #59, opened 2022-12-08
A user can mark a booking as extended

bryangaledxw:
# Changes in this PR

* A user can now extend an active booking, using the action menu on the show booking page

## Screenshots of UI changes

localhost_3000_properties_af1f09fc-9a2a-43fa-a2a9-14ea15b9a0d1_bedspaces_8d2e2951-1a20-4e23-bc27-af1fd2064cd2_bookings_ce6c17ca-e71d-45f7-ab35-b8e8428cc45e_extend_new


bryangaledxw:
[A user can mark a booking as 'extended'](https://trello.com/c/BNxOTVJH/535-a-user-can-mark-a-booking-as-extended)

integration_tests/tests/temporary-accommodation/manage/extension.cy.ts
```diff
@@ -14,6 +14,29 @@ context('Booking extension', () => {
     cy.task('stubAuthUser')
   })
 
+  it('navigates to the booking extension page', () => {
```
pezholio, line 17: This is a strong opinion, loosely held, and don't feel you need to change anything, but I feel like integration tests for small things, like making sure you can navigate to a page, would be better rolled into the main feature, as Cypress tests tend to come with a quite hefty price in terms of run time. It's entirely personal preference though, and I can see both sides of the argument.

e2e/tests/booking.feature
```diff
@@ -35,6 +35,22 @@ Feature: Manage Temporary Accommodation - Booking
         And I attempt to mark the booking as arrived with required details missing
         Then I should see a list of the problems encountered marking the booking as arrived
 
+    Scenario: Extending a booking
```
pezholio, line 38: Respect for keeping the e2e tests up to date. We've got a bit lazy in this regard

server/services/bookingService.ts
```diff
@@ -144,6 +144,18 @@ export default class BookingService {
...
+      const collatedNotes = booking.extensions
+        .map(extension => extension.notes)
+        .filter(notes => notes?.length > 0)
+        .join('\n\n')
+
+      rows.push({
+        key: this.textValue('Extension notes'),
+        value: this.htmlValue(formatLines(collatedNotes)),
+      })
+    }
```
pezholio, line 157: Again, this is not somehting to act on now, but I've started to move this kind of UI-first work to util functions, which we can then use in the views themselves. It makes it easier to test and seperates the data and presentation later a bit better.

pezholio (approved):
LGTM - just some random musings in the comments

integration_tests/tests/temporary-accommodation/manage/extension.cy.ts
bryangaledxw: Hmm. My thinking was, do this bit of navigation once in it's own test, then I can be quick/lazy in the rest of the tests and just jump directly to the page. Though if there's a clear happy path to roll this into....

e2e/tests/booking.feature
bryangaledxw: tbh it's causing some problems! I've accidentally pushed us into this odd situation where reference data comes from the API, *but* the UI still needs it's own copy of that data for the E2E tests

I was taking a peek at your E2E tests just now, and I saw you have 
```
  const lostBedReasonId = Cypress.env('lost_bed_reason_id') || throwMissingCypressEnvError('lost_bed_reason_id')
```

to get round this, but I don't think I want that everywhere we're selecting from a list of options, and I don't think I want to go back to my original approach of inspecting the page to find an ID for a well-known name. I'm wondering if I need to spend some time on a generic approach to essentially derive a set of reference data from a dropdown/radio list

server/services/bookingService.ts
bryangaledxw: Ah, yep, saw that. Yeah, may well follow that...

PR #61, opened 2022-12-09
Remove Approved Premises-only reference data

bryangaledxw:
Remove Approved Premises-only departure reasons and move on categories. These were causing occasional E2E test failures


danlivings-dxw (approved):
LGTM

PR #64, opened 2022-12-12
Workflow to update the API models

tahb:
Copy the AP approach[1] to semi-automate the updates to API models made upstream.

As a workflow dispatch this will need to be run manually via github actions.

[1] https://github.com/ministryofjustice/hmpps-approved-premises-ui/commit/f957fa0bd5cc8e2a6758afb5dd455f310f999b2a

### DIFF
--- a/.github/workflows/generate-types.yml
+++ b/.github/workflows/generate-types.yml
@@ -1,0 +1,28 @@
+name: Generate Types
+
+on:
+  workflow_dispatch:
+
+env:
+  NODE_ENV: test
+  API_CLIENT_ID: approved-premises
+  API_CLIENT_SECRET: clientsecret
+
+jobs:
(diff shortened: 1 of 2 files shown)


bryangaledxw:
For API changes that require further code changes - new required fields for instance, is the idea that we take the automatically created branch as the starting point and add to it as needed?

tahb:
@bryangaledxw Yes I believe so. As in my [recent example](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/65) I ran the script and then had to do another edit to the factory to make it pass. I think there is value in addressing API changes in isolation like this as it avoids these fixes being woven through our main pull requests.

bryangaledxw (approved):

PR #65, opened 2022-12-12
Chore/update model types

tahb:
I ran script/generate-types and then manually udpated the factory all in an effort to fix the e2e tests so we can deploy.

I've also tried to help us automate this a bit with the AP workflow to automatically start the PR https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/64

### DIFF
--- a/server/testutils/factories/booking.ts
+++ b/server/testutils/factories/booking.ts
@@ -1,13 +1,13 @@
-import { Factory } from 'fishery'
 import { faker } from '@faker-js/faker/locale/en_GB'
 import { addDays } from 'date-fns'
+import { Factory } from 'fishery'
 
 import type { Booking } from '@approved-premises/api'
-import arrivalFactory from './arrival'
-import departureFactory from './departure'
-import confirmationFactory from './confirmation'
-import personFactory from './person'
 import { DateFormats } from '../../utils/dateUtils'
+import arrivalFactory from './arrival'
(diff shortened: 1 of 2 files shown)


server/testutils/factories/booking.ts
```diff
@@ -1,13 +1,13 @@
...
+import { Factory } from 'fishery'
 
 import type { Booking } from '@approved-premises/api'
+import { DateFormats } from '../../utils/dateUtils'
 import arrivalFactory from './arrival'
-import departureFactory from './departure'
 import confirmationFactory from './confirmation'
+import departureFactory from './departure'
 import personFactory from './person'
-import { DateFormats } from '../../utils/dateUtils'
```
tahb, line 10: As mentioned in my commit, my editor is doing some automatic linting here. Happy to remove if it's not the way? looks Looks like it's sorting it alphabetically and by case.

richpjames: I like this order of import but not sure what Bryan thinks

richpjames (approved):
LGTM to me. 
I don't think the ordering of imports will be a problem

server/testutils/factories/booking.ts
tahb: I shall merge this to see if I can deploy and beg forgiveness from you @bryangaledxw, let me know if I need to turn this off.

PR #68, opened 2022-12-15
Chore/document release process

tahb:
As discussed with the team at the office day. The draft process has now had a bit of time to sit and is not published with one amendment: Product review should start as opt-in during the planning stage to not overly hold up delivery. 

Once we're happy here I'll propose similar changes to AP UI and the API.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -11,3 +11,35 @@
 ### Before
 
 ### After
+
+# Release checklist
+
+As part of our continuous deployment strategy we must ensure that this work is
+ready to be released at any point. Before merging to `main` we must first
+confirm:
+
+## Pre merge checklist
+
+- [ ] Have all changes to any dependencies been deployed to both `preprod` and
+    `production` in a backwards compatible way? (This includes our API,
+    infrastructure, third-party integrations and any secrets or environment variables)
+- [ ] Has a required data migration been included in this change or been done in
+    advance?
+
+## Post merge checklist
+
+Once we've merged we now need to release this through to production before
+considering it done.
+
+[Find the build-and-deploy job in CircleCI](https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-temporary-accommodation-ui)
+and work through the following steps:
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -70,3 +70,7 @@ script/local_e2e
 ```
 
 Note: This requires `ap-tools` to be installed (<https://github.com/ministryofjustice/hmpps-approved-premises-tools>)
+
+## Release process
+
+Our release process aligns with the other CAS teams and as such [lives in Confluence](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process)
```
edavey: Perhaps the README should also link to [.github/PULL_REQUEST_TEMPLATE.md](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/68/commits/ce41a72cb84b8f6dedbf71f06e82efcf49f13c33#diff-18813c86948efc57e661623d7ba48ff94325c9b5421ec9177f724922dd553a35) so that more detail is readily discoverable from within the repo itself?

There's always a risk when linking to other resources link Confluence that the viewer won't have access or the wiki page is moved/removed...

edavey (approved):
👍 just a tiny suggestion about making [.github/PULL_REQUEST_TEMPLATE.md](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/68/commits/438c184577bba3378947b63e69b9cab3d096de95#diff-18813c86948efc57e661623d7ba48ff94325c9b5421ec9177f724922dd553a35) more visible

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -11,3 +11,35 @@
 ### Before
 
 ### After
+
+# Release checklist
```
pezholio, line 15: I'm not an enormous fan of all this detail in PR templates, it feels a bit messy and I always tend to delete a bunch of it as it's repetitive. Not for now, but I do wonder if there's a tider way of doing this with actions or similar?

pezholio (approved):
I'm happy with this. I've made a comment about PR templates, but I don't think it's a blocker. I think Ed's point about linking to the template from the Readme is a good one.

PR #69, opened 2022-12-16
Add a dashboard as the home page and allow users to click through to 'Manage properties'

bryangaledxw:
# Changes in this PR

* This PR adds a dashboard to the TA service. When the user first arrives at the service, they will see this dashboard, rather than be redirected to the premises listing

## Screenshots of UI changes

localhost_3000_


bryangaledxw:
[Add a dashboard as the home page and allow users to click through to 'Manage properties'](https://trello.com/c/tN51Bzqo/637-add-a-dashboard-as-the-home-page-and-allow-users-to-click-through-to-manage-properties)

pezholio (approved):
LGTM  - we should really do something similar too!

PR #86, opened 2023-01-09
Increase timeout when downloading report in E2E test

bryangaledxw:
The API call behind this seems to sometimes take longer than the default 4 seconds, so we increase the timeout to fix an intermittent test failure


tahb (approved):
👍

PR #88, opened 2023-01-10
Fix report download timeout

bryangaledxw:
Previously, we were only setting a timeout for the "specific region" report E2E test, causing a test failure


tahb (approved):
👍

PR #94, opened 2023-01-11
Fix PDU data

bryangaledxw:
We used the wrong data source for the PDUs, so we change this to the correct data


pezholio (approved):
LGTM - sorry, should have checked over the data!

pezholio (approved):

PR #97, opened 2023-01-12
Chore/deploy to prod

tahb:
# Context

https://trello.com/c/EnjCxzGs/12-set-up-production

Add the Helm deploy charts to configure the deployed app and CircleCI to allow us to deploy.

# Release checklist

confirm:

## Pre merge checklist

- [x] Have all changes to any dependencies been deployed to both `preprod` and
- [x] Has a required data migration been included in this change or been done in
    advance?

## Post merge checklist

considering it done.

fix it as soon as possible.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -141,22 +141,20 @@ workflows:
           requires:
             - request-preprod-approval
 
-  #      - request-prod-approval:
-  #          type: approval
-  #          requires:
-  #            - deploy_preprod
-  #      - hmpps/deploy_env:
-  #          name: deploy_prod
-  #          env: "prod"
-  #          jira_update: true
-  #          jira_env_type: production
-  #          slack_notification: true
-  #          slack_channel_name: << pipeline.parameters.releases-slack-channel >>
-  #          context:
-  #            - hmpps-common-vars
-  #            - approved-premises-ui-prod
-  #          requires:
-  #            - request-prod-approval
+      - request-prod-approval:
+          type: approval
+          requires:
+            - deploy_preprod
+      - hmpps/deploy_env:
+          name: deploy_prod
+          env: "prod"
+          slack_notification: true
+          slack_channel_name: << pipeline.parameters.releases-slack-channel >>
+          context:
+            - hmpps-common-vars
+            - hmpps-temporary-accommodation-ui-prod
+          requires:
+            - request-prod-approval
 
   security:
     triggers:
--- a/helm_deploy/values-prod.yaml
+++ b/helm_deploy/values-prod.yaml
@@ -1,5 +1,5 @@
 ---
-# Per environment values which override defaults in approved-premises-ui/values.yaml
+# Per environment values which override defaults in hmpps-temporary-accommodation-ui/values.yaml
 
 generic-service:
   ingress:
--- a/helm_deploy/values-prod.yaml
+++ b/helm_deploy/values-prod.yaml
@@ -1,0 +1,16 @@
+---
+# Per environment values which override defaults in approved-premises-ui/values.yaml
+
+generic-service:
+  ingress:
+    hosts:
+      - temporary-accommodation.hmpps.service.justice.gov.uk
+    contextColour: green
+    tlsSecretName: hmpps-temporary-accommodation-prod-cert
+
+  env:
+    ENVIRONMENT: prod
+    APPROVED_PREMISES_API_URL: 'https://approved-premises-api.hmpps.service.justice.gov.uk'
+    INGRESS_URL: 'https://temporary-accommodation.hmpps.service.justice.gov.uk'
+    HMPPS_AUTH_URL: 'https://sign-in.hmpps.service.justice.gov.uk/auth'
+    TOKEN_VERIFICATION_API_URL: 'https://token-verification-api.prison.service.justice.gov.uk'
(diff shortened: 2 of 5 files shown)


helm_deploy/values-prod.yaml
```diff
@@ -0,0 +1,19 @@
+---
+# Per environment values which override defaults in approved-premises-ui/values.yaml
```
bryangaledxw: Copy and paste error in comment? `approved-premises-ui` should be temporary-accommodation-ui`

tahb: Yeah good spot, that should be in all 4 helm charts so I'll add another commit to rename them 👍

tahb:
We now have [authority to operate](https://trello.com/c/MU5j9F5L) so this is ready to proceed.

.circleci/config.yml
```diff
@@ -141,22 +141,20 @@ workflows:
...
-  #            - request-prod-approval
+      - request-prod-approval:
+          type: approval
+          requires:
+            - deploy_preprod
+      - hmpps/deploy_env:
+          name: deploy_prod
+          env: "prod"
+          slack_notification: true
+          slack_channel_name: << pipeline.parameters.releases-slack-channel >>
```
edavey, line 152: Is it worth backfilling these slack setting to pre-prod to verify that we've got the wiring and permissions working? (Assuming we do feel that slack notifications for pre-prod as well as prod are of value...)

helm_deploy/values-prod.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+  env:
+    ENVIRONMENT: prod
+    APPROVED_PREMISES_API_URL: 'https://approved-premises-api.hmpps.service.justice.gov.uk'
+    INGRESS_URL: 'https://temporary-accommodation.hmpps.service.justice.gov.uk'
+    HMPPS_AUTH_URL: 'https://sign-in.hmpps.service.justice.gov.uk/auth'
+    TOKEN_VERIFICATION_API_URL: 'https://token-verification-api.prison.service.justice.gov.uk'
+
+generic-prometheus-alerts:
+  alertSeverity: digital-prison-service
```
edavey: in pre-prod we have `digital-prison-service-dev` -- I'm guessing you've seen guidance / examples of using this dev-less setting in production?

Looking at Make a Recall Decision, I see they seem to have defined their own profile which has the advantage of having a super-descriptive name: https://github.com/ministryofjustice/make-recall-decision-ui/blob/38428206219454c674d737d58f7ac0c66df5cbd3/helm_deploy/values-prod.yaml#L151-L152

.circleci/config.yml
edavey: I note that MRD have configured a slack action for build failure: https://github.com/ministryofjustice/make-recall-decision-ui/blob/25b3e73462e577516a79c6014a9a113033b8d842/.circleci/config.yml#L33-L61

might be worth copying, or making a card in the backlog to copy...

edavey (approved):
This looks correct. Just a few questions / suggestions. Nothing to block merging.

.circleci/config.yml
tahb: Yes it's true we haven't tested that slack notification. I was thinking that because we have [CI job failure events coming through on Slack](https://mojdt.slack.com/archives/C048BJS7S2F/p1675234943024679) we have enough of a signal for non production environments. For production I was thinking that we want something even more shouty and that might alert the wider team of a potential issue. 

I'll add a link to what MRD have done in case we feel the need to reach for it.

helm_deploy/values-prod.yaml
tahb: My understanding is that [this value is for routing slack notifications](https://github.com/ministryofjustice/hmpps-helm-charts/blob/ce5cef6065ad33448b2606a82fe02e8c1bf357e9/charts/generic-prometheus-alerts/README.md) and this value will send a slack alert to 'dps_alerts', I don't think we can send multiple values to this. My general understanding of Prometheus is that it's job is to handle platform level alerting such as a database being unavailable. For those types of alert our team can't really do too much so I figured dps_alerts would be the best place for it to bubble up to.

The documentation does say to consult the platform so I will double check this, good shout 👍

tahb: [I've asked the CP team](https://mojdt.slack.com/archives/C57UPMZLY/p1675271360663659)

.circleci/config.yml
tahb: On further look perhaps that existing notification is not telling us of failed deploys..

tahb: Okay it looks like that slack event is coming from this line https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/f8e442ad46eb65de49048317cbd62d20d2a6f1f3/.circleci/config.yml#L41 and a few others.

I think we ship this to get production notifications going the same way and sweep back to add them for prod. I'll put something in our backlog.

helm_deploy/values-prod.yaml
tahb: From a chat there I think this value essentially does nothing for us right now.

We use Sentry for application errors and the platform team have their own alerting set up to keep an eye on what's inside the cluster. 

Should we want to add more custom alerts or switch away from Sentry we could [set up custom alerts](https://user-guide.cloud-platform.service.justice.gov.uk/documentation/monitoring-an-app/how-to-create-alarms.html#creating-your-own-custom-alerts) and use this value again.

PR #104, opened 2023-01-19
Set up dev environment for auditing

bryangaledxw:
We ensure the necessary environment variables are set to start logging audit data on our development environment


bryangaledxw:
The main thing I'm nervous about with this PR is if the values under `namespace_secrets:` in `values.yaml` and `values-dev.yaml` will get merged as I'm expecting, or if the latter will just stomp the former. Q for someone who know this stuff better than I do

pezholio (approved):
LGTM 👍

tahb:
I think this will handle as you expect.

For Helm I believe that values.yml contain the defaults, one or more files can then provided via the `--values` flag can be passed in to override ([MoJ have an orb that grabs our values-x.yml file for use in this way](https://github.com/ministryofjustice/hmpps-circleci-orb/blob/main/src/scripts/deploy_env.sh#L29)). Or, you can use `--set` to override both.

PR #108, opened 2023-01-23
Use username rather than UUID in audit events

bryangaledxw:
Following discussions with the audit team, we send the username rather than the UUID in audit events


tahb (approved):
LGTM.

[Record of the audit teams decision around data handling can be found in this Slack thread](https://mojdt.slack.com/archives/C028V9TS295/p1674489555007459).

PR #112, opened 2023-01-25
Pass request into RestClient

bryangaledxw:
This is a refactoring PR as part of our work to limit users to properties of their own region. Our intent is to store the user region in the session, which we will then pass back to the API on each call, within an `X-USER-REGION` header. To do this, we either need to pass an additional argument into our `RestClient` alongside the `token`, or we pass the request object into the `RestClient`, and let the `RestClient` extract both the token from `req.user` and later the user region `req.session`.

Though a scary number of files have changed, the only "real" change is to `restClient.ts`, and everything else simply flows from there, so we pass `req` through our controllers, services, and clients, rather than `token`


richpjames (approved):
LGTM, agree with the approach here.

pezholio:
I think I’d prefer to see the headers passed in alongside the token. Passing the request object round makes testing harder and it’s also more difficult to see what the intent is

PR #115, opened 2023-01-25
Generate latest shared types

bryangaledxw:
We generate the latest shared types, and make the necessary changes within the Apply feature (currently unchanged since we forked from Approved Premises) to support these types. This is mostly a switch from using Application to ApprovedPremisesApplication


richpjames (approved):
Apologies as I'm missing some context but am I right in thinking we're using the`ApprovedPremisesApplication` in the Temporary Accommodation repo (whilst`TemporaryAccommodationApplication` exists) because so far the Apply journey for TA hasn't been built out and is essentially the same as AP Apply currently?
If so I'm approving with the assumption that `TemporaryAccommodationApplication` will be used when the time is right

bryangaledxw:
Ah, yep, this is very much *get things compiling* against the latest types. I did originally make this PR "strip out AP apply code", but I think I wanted to do the smallest PR that'd work against the new types

PR #124, opened 2023-01-30
Pass call config into RestClient

bryangaledxw:
This is a refactoring PR as part of our work to limit users to properties of their own region. Our intent is to store the user region in the session, which we will then pass back to the API on each call, within an `X-USER-REGION` header. To do this, we need to pass additional information from the request into the RestClient.

To do this, we replace the token that `RestClient` currently takes, with `CallConfig` object, and an `extractCallConfig` helper function used by our controllers to create a `CallConfig` from our request. For now, the `callConfig` only contains the token, but we will expand this in a future PR.

The PR is split into 3 broad steps:
- Controllers passing a callConfig into our services
- Services passing a callConfig into our clients
- Clients passing a callConfig into our RestClient


pezholio (approved):
This looks much better - thanks for taking the time to do this 👍

PR #136, opened 2023-02-02
Reduce duplication in integration test setup

bryangaledxw:
We had a fair amount if duplication in the beforeEach blocks of our integration tests, which the addition of user regions made worse. We now reduce duplication by expanding the responsibilities of `stubActingUser`, which is now renamed to `setupTestUser`.


richpjames (dismissed):
LGTM

richpjames (approved):

PR #139, opened 2023-02-02
Property listing only shows properties for user region(s)

bryangaledxw:
Property filtering is handled API-side, though we add an E2E test to check we are only seeing properties for the user's region.


richpjames (approved):

bryangaledxw:
[Property listing only shows properties for user region(s)](https://trello.com/c/MRs57957/671-property-listing-only-shows-properties-for-user-regions)

patrickjfl (approved):
LGTM 👍

PR #149, opened 2023-02-08
Allow ARK/DOM1 IP addresses to access the service

tahb:
# Changes in this PR

At the moment our users are not using a MOJ VPN and so can't access production.

They could access the Create and Vary a license service which manages its infrastructure similar to us. We can see there that they have a much larger allow list[1]

We can see from the Manage and Supervision team that they have previously checked if the DOM1 network IP addresses will enable most of their users[2], which was confirmed except for Greater Manchester.

The IP addresses that we are adding here are taken from the Delius allow list[3] on the basis that if the user can access Delius, they should also be able to access our service.

We may have to add more IP addresses over time as we learn more about which users are on which network.

[1] https://github.com/ministryofjustice/create-and-vary-a-licence/blob/main/helm_deploy/create-and-vary-a-licence/values.yaml#L74 
[2] https://mojdt.slack.com/archives/C027KSVD0F5/p1634220661003900 
[3] https://github.com/ministryofjustice/hmpps-env-configs/blob/a222488e6e7a3f233f793b80197384b2136b5aee/common-prod/common.tfvars#L413

# Release checklist

confirm:

## Pre merge checklist

    advance?

## Post merge checklist

considering it done.

fix it as soon as possible.

### DIFF
--- a/helm_deploy/hmpps-temporary-accommodation-ui/values.yaml
+++ b/helm_deploy/hmpps-temporary-accommodation-ui/values.yaml
@@ -68,6 +68,11 @@ generic-service:
     cloudplatform-live1-1: '35.178.209.113/32'
     cloudplatform-live1-2: '3.8.51.207/32'
     cloudplatform-live1-3: '35.177.252.54/32'
+    ark-nps-hmcts-ttp1: "195.59.75.0/24"
+    ark-nps-hmcts-ttp2: "194.33.192.0/25"
+    ark-nps-hmcts-ttp3: "194.33.193.0/25"
+    ark-nps-hmcts-ttp4: "194.33.196.0/25"
+    ark-nps-hmcts-ttp5: "194.33.197.0/25"
 
 generic-prometheus-alerts:
   targetApplication: hmpps-temporary-accommodation-ui


helm_deploy/hmpps-temporary-accommodation-ui/values.yaml
```diff
@@ -68,6 +68,11 @@ generic-service:
     cloudplatform-live1-1: '35.178.209.113/32'
     cloudplatform-live1-2: '3.8.51.207/32'
     cloudplatform-live1-3: '35.177.252.54/32'
+    ark-nps-hmcts-ttp1: "195.59.75.0/24"
+    ark-nps-hmcts-ttp2: "194.33.192.0/25"
+    ark-nps-hmcts-ttp3: "194.33.193.0/25"
+    ark-nps-hmcts-ttp4: "194.33.196.0/25"
```
tahb, line 74: We've verified that our first user is on DOM within this range so it should solve the immediate issue of smoke testing prod.

edavey: for maintainability and clarity would it be worth incorporating "DOM1" into this somehow e.g. within the keys like:

```
ark-nps-hmcts-dom1-ttp1: "195.59.75.0/24"
...
```

edavey (approved):
this looks correct to me, thks.

helm_deploy/hmpps-temporary-accommodation-ui/values.yaml
tahb: I wasn't sure if the keys had to match something downstream given that's how Create and Vary have done it. I'll see if we can set our own names in the next batch 👍

PR #154, opened 2023-02-09
Fix premises alias in E2E tests

bryangaledxw:
We previously changed our "I'm viewing an existing premises" E2E test step to alias a single-element array containing a premises as 'premisesList', when previously it had been aliasing this single premises as 'premises'.

This broke our bedspace and bookings E2E tests, as these depended on a premises being aliased to 'premises'. We now alias the single element from 'premisesList' as 'premises', restoring our bedspace and bookings tests


tahb (approved):
LGTM

PR #180, opened 2023-02-23
Fix Sentry by adding integrations

tahb:
# Changes in this PR

I had hoped that configuring tracing would start to send transactions through to Sentry. It appears not, even on production where sampling is 1. I've not connected this locally to Sentry and can see transactions coming through!

Now we copy the rest of the Make a Recall set up[1] that adds integrations which, according to the commends, "enables HTTP call tracing".

We filter out requests to /assets to prevent spam and anonymise the username.

[1] https://github.com/ministryofjustice/manage-recalls-ui/blob/c3b4fd4486b26da47c086778fdc89e7a6befcde0/server/middleware/setUpSentry.ts

## Screenshots of UI changes

(Note the usernames that appear in earlier requests and then are filtered out in later requests after noticing them)

Screenshot 2023-02-24 at 12 18 38

# Release checklist

confirm:

## Pre merge checklist

    advance?

## Post merge checklist

considering it done.

fix it as soon as possible.

### DIFF
--- a/server/middleware/setUpSentry.ts
+++ b/server/middleware/setUpSentry.ts
@@ -1,24 +1,38 @@
 import * as Sentry from '@sentry/node'
-import '@sentry/tracing'
+import * as Tracing from '@sentry/tracing'
 import express from 'express'
 import applicationVersion from '../applicationVersion'
 import config from '../config'
 
 export function setUpSentryRequestHandler(app: express.Express): void {
   if (config.sentry.dsn) {
+    // Prevent usernames which are PII from being sent to Sentry
+    // https://docs.sentry.io/platforms/python/guides/logging/data-management/sensitive-data/#examples
+    const anonymousId = Math.random().toString()
+    Sentry.setUser({ id: anonymousId, username: anonymousId })
+
     Sentry.init({
       dsn: config.sentry.dsn,
       environment: config.environment,
       release: applicationVersion.gitRef,
+      integrations: [
+        // enable HTTP calls tracing
+        new Sentry.Integrations.Http({ tracing: true }),
+        // enable Express.js middleware tracing
+        new Tracing.Integrations.Express({ app }),
+      ],
       tracesSampler: samplingContext => {
         const transactionName = samplingContext?.transactionContext?.name
-
-        if (transactionName?.includes('ping') || transactionName?.includes('health')) {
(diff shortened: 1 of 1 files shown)


server/middleware/setUpSentry.ts
```diff
@@ -1,15 +1,23 @@
 import * as Sentry from '@sentry/node'
-import '@sentry/tracing'
+import * as Tracing from '@sentry/tracing'
 import express from 'express'
 import applicationVersion from '../applicationVersion'
 import config from '../config'
 
 export function setUpSentryRequestHandler(app: express.Express): void {
   if (config.sentry.dsn) {
+    const router = express.Router()
```
bryangaledxw: I think you'll need to call  `app.use(router)` for this to do anything. In make-recalls-ui, setupSentry returns the router, and `app.use(setUpSentry())` is called in app.ts

bryangaledxw: Also, I expect the call to `router.use(Sentry.Handlers.tracingHandler())` in make-recalls-ui is significant

tahb: Ah I didn't see this feedback before pushing my latest changes. I did end up removing this line 👍 

And yes that line was very significant as it turns out 😆

server/middleware/setUpSentry.ts
```diff
@@ -1,24 +1,38 @@
...
+import * as Tracing from '@sentry/tracing'
 import express from 'express'
 import applicationVersion from '../applicationVersion'
 import config from '../config'
 
 export function setUpSentryRequestHandler(app: express.Express): void {
   if (config.sentry.dsn) {
+    // Prevent usernames which are PII from being sent to Sentry
+    // https://docs.sentry.io/platforms/python/guides/logging/data-management/sensitive-data/#examples
+    const anonymousID = Math.random().toString()
```
bryangaledxw: We're generally using `roomId`, `premisesId` etc., would change to `anonymousId` for consistency

bryangaledxw (approved):
Once incredibly picky point but otherwise looks good

server/middleware/setUpSentry.ts
tahb: Didn't spot that, will change! 👍

PR #184, opened 2023-02-24
Users can mark a bedspace unavailable for bookings

patrickjfl:
# Context

Allows users to void a bedspace, and to view void 'bookings'. See [Trello ticket](https://trello.com/c/aP1llGe6/756-users-can-mark-a-bedspace-as-unavailable-for-bookings) for more information. Much of the work shadows the existing Bookings journey. For now the Voids routes and link within the UI are feature flagged, pending API support.

Also two minor tweaks, one to the .env.example file, and another to the restClient logging.

# Changes in this PR

## Screenshots of UI changes

Screenshot 2023-02-24 at 16 04 58
Screenshot 2023-02-24 at 16 04 43

### Before

Screenshot 2023-02-24 at 16 11 34

### After

Screenshot 2023-02-24 at 16 04 35

# Release checklist

confirm:

## Pre merge checklist

    advance?

## Post merge checklist

considering it done.

fix it as soon as possible.


bryangaledxw:
For the Trello link, you can use the feature in Trello to get a bidirectional link between ticket and PR
image

server/testutils/factories/newLostBed.ts
```diff
@@ -3,24 +3,23 @@ import { faker } from '@faker-js/faker/locale/en_GB'
...
-    startDate: DateFormats.dateObjToIsoDate(startDate),
+    startDate: null,
     'startDate-day': startDate.getDate().toString(),
     'startDate-month': startDate.getMonth().toString(),
     'startDate-year': startDate.getFullYear().toString(),
-    endDate: DateFormats.dateObjToIsoDate(endDate),
+    endDate: null,
     'endDate-day': endDate.getDate().toString(),
     'endDate-month': endDate.getMonth().toString(),
     'endDate-year': endDate.getFullYear().toString(),
```
bryangaledxw: My preference would be go the other way here - keep the factory 'honest' wrt the `NewLostBed` type, and what gets sent from the UI to the API, and do any extra work for turn this into what gets sent from the page to the UI in the test

I've run into factories inherited from CAS1 doing this, and got rid of it because it ended up being something of a tripwire

wiremock/lostBedStubs.ts
```diff
@@ -15,10 +15,41 @@ lostBeds.push({
...
+    jsonBody: lostBedFactory.build(),
   },
 })
-const requiredFields = getCombinations(['startDate', 'endDate', 'numberOfBeds', 'reason', 'referenceNumber'])
+
+lostBeds.push({
+  priority: 99,
+  request: {
+    method: 'GET',
+    urlPathPattern: `/premises/${guidRegex}/lost-beds/${guidRegex}`,
```
bryangaledxw: It feels a little dodgy given that you'd be sticking the `guidRegex` in as an ID, but it seems static paths do work here. See `applicationStubs.ts`

server/components/lostBedInfo.ts
```diff
@@ -0,0 +1,41 @@
+import type { LostBed } from '@approved-premises/api'
+import type { SummaryList } from '@approved-premises/ui'
+import { DateFormats } from '../utils/dateUtils'
+import { formatLines } from '../utils/viewUtils'
+import { lostBedStatusHtml } from '../utils/lostBedUtils'
+
+export default (lostBed: LostBed): SummaryList['rows'] => {
```
bryangaledxw, line 7: Does this go with a macro and changes in nunjucksSetup.ts?  Similar to `booking-info/macro.njk`/`bookingInfo.ts`? Suggest add those all together in one commit

server/services/lostBedService.ts
```diff
@@ -13,11 +13,19 @@ export default class LostBedService {
   async createLostBed(callConfig: CallConfig, premisesId: string, lostBed: NewLostBed): Promise<LostBed> {
     const lostBedClient = this.lostBedClientFactory(callConfig)
 
-    const confirmedLostBed = await lostBedClient.create(premisesId, lostBed)
+    const confirmedLostBed = (await lostBedClient.create(premisesId, lostBed)) as LostBed
```
bryangaledxw: I don't think the cast is needed?

server/services/bookingService.test.ts
```diff
@@ -1,26 +1,35 @@
 import BookingService from './bookingService'
 import BookingClient from '../data/bookingClient'
+import LostBedClient from '../data/lostBedClient'
```
bryangaledxw, line 3: Suggest keep the changes to booking service in their own commit

server/controllers/temporary-accommodation/manage/lostBedsController.test.ts
```diff
@@ -0,0 +1,183 @@
...
+          'startDate-month': 8,
+          'startDate-day': 22,
+          'endDate-year': 2022,
+          'endDate-month': 9,
+          'endDate-day': 22,
+          notes: lostBed.notes,
+          reason: lostBed.reason.id,
+          bedId: lostBed.bed.id,
+          referenceNumber: lostBed.referenceNumber,
+          serviceName: 'temporary-accommodation',
```
bryangaledxw: Can you use the newLostBedFactory, and replace this all with something like:
```
...newLostBed,
...DateFormats.convertIsoToDateAndTimeInputs(lostBed.startDate, 'startDate'),
...DateFormats.convertIsoToDateAndTimeInputs(lostBed.endDate, 'endDate'),
```
?

server/controllers/temporary-accommodation/manage/lostBedsController.test.ts
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        await requestHandler(request, response, next)
+
+        expect(lostBedService.createLostBed).toHaveBeenCalledWith(callConfig, premisesId, {
+          ...request.body,
+          startDate: '2022-08-22',
+          endDate: '2022-09-22',
```
bryangaledxw: ... then this can become something like `expect.objectContaining(newLostBed)`

server/views/temporary-accommodation/components/lost-bed-info/macro.njk
```diff
@@ -0,0 +1,15 @@
+{% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
```
bryangaledxw, line 1: For all the templates, add a newline at the end, and suggest add some more whitespace for readability

server/config.ts
```diff
@@ -49,6 +49,7 @@ export default {
   staticResourceCacheDuration: 20,
   flags: {
     oasysDisabled: process.env.OASYS_DISABLED || false,
+    voidsDisabled: !['local', 'dev', 'test'].includes(process.env.ENVIRONMENT || 'local'),
```
bryangaledxw: Suggest pull out `const environment = process.env.ENVIRONMENT || 'local'`, and use that here and on line 54

cypress_shared/pages/temporary-accommodation/manage/lostBedEditable.ts
```diff
@@ -0,0 +1,30 @@
...
+export default abstract class LostBedEditablePage extends Page {
+  shouldShowDateConflictErrorMessages(): void {
+    ;['startDate', 'endDate'].forEach(field => {
+      cy.get('.govuk-error-summary').should('contain', errorLookups.generic[field].conflict)
+      cy.get(`[data-cy-error-${field}]`).should('contain', errorLookups.generic[field].conflict)
+    })
+  }
+
+  protected completeEditableForm(newOrUpdateLostBed: NewLostBed, roomId: string): void {
+    this.getHiddenInputByNameAndSetValue('bedId', roomId)
```
bryangaledxw: I feel like in principle this shouldn't be necessary? Everything on the page is coming from our own faked API, so we should be able control the contents of the hidden for via that?

server/services/lostBedService.ts
```diff
@@ -13,11 +13,19 @@ export default class LostBedService {
   async createLostBed(callConfig: CallConfig, premisesId: string, lostBed: NewLostBed): Promise<LostBed> {
```
bryangaledxw: Suggest rename to "create" for consistency

server/views/temporary-accommodation/lost-beds/new.njk
```diff
@@ -0,0 +1,29 @@
...
+        <h2>Bedspace reference</h2>
+        <p>{{ room.name }}</p>
+        <h2>Property address</h2>
+        <p>{{ premises.addressLine1 }}<br/>{{ premises.postcode }}</p>
+    </div>
+    <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+            <form action="{{ paths.lostBeds.create({ premisesId: premises.id, roomId: room.id }) }}" method="post">
+                <input type="hidden" name="_csrf" value="{{ csrfToken }}"/>
+                <input type="hidden" name="bedId" value="{{ room.id }}"/> {% include "./_editable.njk" %}
```
bryangaledxw: Is this meant to be `value="{{ room.beds[0].id }}"`?

So I understand this correctly, this is so that in the create endpoint, we don't have to fetch the room again *just* to get the bed ID (this is what BookingsController does, but I think your solution here is good)

integration_tests/tests/temporary-accommodation/manage/lostBed.cy.ts
```diff
@@ -0,0 +1,253 @@
...
+    cy.task('stubSingleRoom', { premisesId: premises.id, room })
+
+    // When I visit the void a bedspace page
+    cy.task('stubLostBedReferenceData')
+    const page = LostBedNewPage.visit(premises.id, room.id)
+
+    // And I fill out the form
+    const lostBed = lostBedFactory.build({
+      bed: bedFactory.build({
+        id: room.id,
```
bryangaledxw: My instinct is something's gone amiss if we're assigning the room ID to the bed ID. Is this fixed by the change I suggested to `new.njk`?

bryangaledxw (commented):
All looks like really good thorough stuff! So please don't take all my comments as a bad sign!

The only real technical thing I'm worried about is in the integration test - needing to set the hidden field in the integration test, and setting the bed ID to the room ID in the when building a lost bed. I'm hoping this goes away with the change I suggested in new.njk.

Other than that, I'd maybe split the commits a bit differently. I'd really like to see the Lost Bed Info component come in in it's own commit, so the code, the nunjucks template, and the nunjucks setup are self contained. And maybe do a commit each for the changes in BookingService and LostBedService

And I suggest in the controller commit, don't include the route changes, or the changes to show.njk in that. I'd instead put those in with the feature flag commit. So there's one commit adding the controller and its dependencies, and another commit actually wiring it up to the rest of the app

Also, for the unrelated tweaks, I think it's best to have all those first and get them out the way (and I'd put deleting `api.ts` in its own commit)

Last minor things - there's a weird character in the commit message for the lost beds controller commit. And there's a stray period in the title of the integration tests commit

patrickjfl:
[Users can mark a bedspace as unavailable for bookings](https://trello.com/c/aP1llGe6/756-users-can-mark-a-bedspace-as-unavailable-for-bookings)

server/views/temporary-accommodation/lost-beds/show.njk
```diff
@@ -0,0 +1,37 @@
...
+      classes: 'govuk-heading-l'
+    },
+    items: actions
+  }) }}
+  <div class="location-header">
+    <h2>Bedspace reference</h2>
+    <p>{{ room.name }}</p>
+    <h2>Property address</h2>
+    <p>{{ premises.addressLine1 }}<br/>{{ premises.postcode }}</p>
+  </div>
```
bryangaledxw: You may well already have this in mind, but let's replace this with the location header component once you've rebased this branch

server/testutils/factories/newLostBed.ts
patrickjfl: yep ok

wiremock/lostBedStubs.ts
patrickjfl: ah ok! I wasn't sure, I think different files are doing different things.

server/components/lostBedInfo.ts
patrickjfl: done

server/services/lostBedService.ts
patrickjfl: 👍

server/config.ts
patrickjfl: good call

cypress_shared/pages/temporary-accommodation/manage/lostBedEditable.ts
patrickjfl: hmmm yeah, not sure what I was trying to achieve there. I guess that was when I was confused about bedIds and roomIds. Removed now anyway 👍

server/views/temporary-accommodation/lost-beds/new.njk
patrickjfl: ah yes good spot!

Yes, that and to maintain consistency in handling the form data. I think it was @richpjames idea actually, so props to him...

integration_tests/tests/temporary-accommodation/manage/lostBed.cy.ts
patrickjfl: yes, good spot

server/views/temporary-accommodation/lost-beds/show.njk
patrickjfl: yep 👍

server/@types/shared/models/LostBed.ts
```diff
@@ -2,18 +2,17 @@
 /* tslint:disable */
 /* eslint-disable */
```
bryangaledxw: Suggest leave this file unlinted, otherwise we'll get spurious changes next time we regenerate types

server/@types/shared/models/LostBedReason.ts
```diff
@@ -3,9 +3,8 @@
 /* eslint-disable */
 
 export type LostBedReason = {
```
bryangaledxw: Suggest leave this file unlinted, otherwise we'll get spurious changes next time we regenerate types

integration_tests/tests/temporary-accommodation/manage/lostBed.cy.ts
```diff
@@ -6,7 +6,6 @@ import newLostBedFactory from '../../../../server/testutils/factories/newLostBed
 import BedspaceShowPage from '../../../../cypress_shared/pages/temporary-accommodation/manage/bedspaceShow'
 import Page from '../../../../cypress_shared/pages/page'
 import LostBedNewPage from '../../../../cypress_shared/pages/temporary-accommodation/manage/lostBedNew'
```
bryangaledxw, line 8: Should the changes in this file be with the integration test commit?

integration_tests/tests/temporary-accommodation/manage/lostBed.cy.ts
```diff
@@ -163,9 +160,7 @@ context('Lost bed', () => {
 
     // And I fill out the form with dates that conflict with an existing lost bed
     const lostBed = lostBedFactory.build({
-      bed: bedFactory.build({
-        id: room.id,
-      }),
+      bedId: room.id,
```
bryangaledxw: Should this and the below change be `bedId: room.beds[0].id`?

e2e/tests/lostBed.feature
```diff
@@ -0,0 +1,18 @@
...
+        Given I'm marking a bedspace as void
+        And I create a void booking with all necessary details
+        Then I should see a confirmation for my new void booking
+
+    Scenario: Showing void booking creation errors
+        Given I'm marking a bedspace as void
+        And I attempt to mark a bedspace as void with required details missing
+        Then I should see a list of the problems encountered voiding the bedspace
+
+
```
bryangaledxw: Suggest tidy so it ends with a newline

cypress_shared/pages/temporary-accommodation/manage/bedspaceShow.ts
```diff
@@ -65,6 +65,7 @@ export default class BedspaceShowPage extends Page {
     cy.get('tr')
       .contains('Void')
       .parent()
+      .parent()
```
bryangaledxw, line 68: I think this should be part of the previous commit?

e2e/tests/stepDefinitions/manage/lostBed.ts
```diff
@@ -0,0 +1,56 @@
+import { Given, Then } from '@badeball/cypress-cucumber-preprocessor'
+import Page from '../../../../cypress_shared/pages/page'
+import BedspaceShowPage from '../../../../cypress_shared/pages/temporary-accommodation/manage/bedspaceShow'
+import LostBedNewPage from '../../../../cypress_shared/pages/temporary-accommodation/manage/lostBedNew'
+import newLostBedFactory from '../../../../server/testutils/factories/newLostBed'
+import lostBedFactory from '../../../../server/testutils/factories/lostBed'
+import LostBedShowPage from '../../../../cypress_shared/pages/temporary-accommodation/manage/lostBedShow'
+
+Given("I'm marking a bedspace as void", () => {
```
bryangaledxw: Appreciate I've not been fully consistent, but I think let's standardise on `Given("I mark a bedspace as void")` etc. for how we word these

server/controllers/temporary-accommodation/manage/lostBedsController.ts
```diff
@@ -0,0 +1,89 @@
...
+      const room = await this.bedspacesService.getRoom(callConfig, premisesId, roomId)
+
+      const lostBed = await this.lostBedsService.find(callConfig, premisesId, lostBedId)
+
+      return res.render('temporary-accommodation/lost-beds/show', {
+        premises,
+        room,
+        lostBed,
+        actions: null,
+        allStatuses,
```
bryangaledxw, line 84: I think this is unused? I can't see that the template does anything with this

server/utils/lostBedUtils.ts
```diff
@@ -0,0 +1,18 @@
+import type { TemporaryAccommodationLostBed } from '@approved-premises/api'
+
+export const allStatuses: Array<{ name: string; id: TemporaryAccommodationLostBed['status']; tagClass: string }> = [
```
bryangaledxw: I think right now this isn't used, and the data here gets duplicated in `lostBedInfo.ts`. I'd rather `lostBedStatusHtml` work the same as `formatStatus` elsewhere, and this just take the status ID string, then lookup human-readable string/colour

server/utils/lostBedUtils.ts
```diff
@@ -0,0 +1,18 @@
...
+    tagClass: 'govuk-tag--green',
+  },
+  {
+    name: 'Cancelled',
+    id: 'cancelled',
+    tagClass: 'govuk-tag--red',
+  },
+]
+
+export const lostBedStatusHtml = (status: string, colour: string): string => {
```
bryangaledxw: Heads up that I've renamed similar functions to `statusTag`, to match CAS1 naming of equivalent functions we've pulled over

server/services/bookingService.ts
```diff
@@ -50,28 +54,51 @@ export default class BookingService {
...
+    const lostBedClient = this.lostBedClientFactory(callConfig)
+    const lostBeds = await (
+      await lostBedClient.allLostBedsForPremisesId(premisesId)
+    ).filter(lostBed => lostBed.status === 'active')
+
     const bedId = room.beds[0].id
 
-    return bookings
-      .filter(booking => booking?.bed.id === bedId)
+    const allReservations = [...bookings, ...lostBeds]
```
bryangaledxw: Feel free to reject this, but I'm wondering if we can lose the need for `isBooking` - it makes me slightly uneasy that we know what's a void and what's a booking, we then lose that information by putting them in an array together, and we then derive it again by being lucky that they don't share a particular field.

We could do something like -

```
const foo1 = bookings.map(b => {sortingValue: v.arrivalDate, rows: [...]}
const foo2 = voids.map(v => {sortingValue: v.startDate, rows: [...]}

return [...foo1, foo2].sort(...).map(bookingOrVoid => bookingOrVoid.rows)
```

We then don't rely on being able to distinguish between voids and bookings in a roundabout way, and we don't need to do `isBooking` for each row

bryangaledxw (commented):
Good stuff on the E2E tests. A few more minor comments. Main thing is on `lostBedStatusHtml`. I think let's have that work the same as similar functions elsewhere - apologies I missed that first time few. I also want to flag my comment on `lostBedController.test` from my first review, about dodging the need to manually specify the date components, I think maybe that go lost with stuff moving round

integration_tests/tests/temporary-accommodation/manage/lostBed.cy.ts
patrickjfl: yes, good spot

e2e/tests/stepDefinitions/manage/lostBed.ts
patrickjfl: ok - as you say there are inconsistencies with other scenario descriptions but we can deal with those later.

server/controllers/temporary-accommodation/manage/lostBedsController.ts
patrickjfl: yes I think I was trying to be clever and get ahead of the game for the edit/cancel work but as you say, not necessary here

server/utils/lostBedUtils.ts
patrickjfl: generally agree, although the problem with changing lostBedStatusHtml is that it's an unusual case - on the view void booking page we want to render 'Active' or 'Cancelled' to the user, whereas on the view bedspace page, if it's an active void booking we want to render 'Void' to the user

patrickjfl: I've followed the pattern but added an optional parameter and if/else for the view bedspace screen case

bryangaledxw: Is this two functions then? One that does a similar mapping from status ID to HTML as elsewhere. Then a function in `viewUtils.ts` for when we just want to turn some string and some colour into HTML, and isn't void-specific?

server/controllers/temporary-accommodation/manage/lostBedsController.test.ts
patrickjfl: yes, good shout

server/services/bookingService.ts
patrickjfl: yes, good idea. Have implemented

server/utils/bookingUtils.test.ts
```diff
@@ -1,15 +1,15 @@
...
   deriveBookingHistory,
   formatStatus,
   getLatestExtension,
   shortenedOrExtended,
 } from './bookingUtils'
+import bookingFactory from '../testutils/factories/booking'
+import paths from '../paths/temporary-accommodation/manage'
+import arrivalFactory from '../testutils/factories/arrival'
+import departureFactory from '../testutils/factories/departure'
+import extensionFactory from '../testutils/factories/extension'
```
bryangaledxw, line 12: I'd either get rid of this, or move it into a "Sort imports" commit. Also, the commit message still has reference to the type guard util that's no longer here

server/services/bookingService.ts
```diff
@@ -50,31 +54,56 @@ export default class BookingService {
...
-          DateFormats.convertIsoToDateObj(a.arrivalDate).getTime()
-        )
-      })
-      .map(booking => {
-        return [
-          this.textValue(booking.person.crn),
-          this.textValue(DateFormats.isoDateToUIDate(booking.arrivalDate, { format: 'short' })),
-          this.textValue(DateFormats.isoDateToUIDate(booking.departureDate, { format: 'short' })),
-          this.htmlValue(formatStatus(booking.status)),
+    const sortedBookings = bookings
```
bryangaledxw: Is there a better name for `sortedBookings` and `sortedLostBeds`, since these aren't sorted?

server/utils/lostBedUtils.ts
```diff
@@ -0,0 +1,22 @@
...
+    tagClass: 'govuk-tag--blue',
+  },
+  {
+    name: 'Cancelled',
+    id: 'cancelled',
+    tagClass: 'govuk-tag--red',
+  },
+]
+
+export function statusTag(statusId: LostBed['status'], viewBedspacePage?: boolean) {
```
bryangaledxw: Maybe change `viewBedspacePage: boolean` to something like `context: 'bookingsAndVoid' | 'voidsOnly'`, so this util doesn't have specific knowledge of a particular page, and give some description as to why there's this variation

cypress_shared/pages/temporary-accommodation/manage/bedspaceShow.ts
```diff
@@ -80,4 +101,8 @@ export default class BedspaceShowPage extends Page {
         cy.get('a').contains('View').click()
       })
   }
+
+  clickLostBedLink(premisesId: string, roomId: string, lostBedId: string): void {
+    cy.get(`[href="${paths.lostBeds.show({ premisesId, roomId, lostBedId })}"]`).click()
```
bryangaledxw, line 107: Consider changing so it's navigating by the link text? I feel like part of the test is "does the text as the user sees it go to the right place"

bryangaledxw: Ah, there might not be a nice way of doing that, right, since you don't have a CRN like with bookings?

e2e/tests/stepDefinitions/manage/lostBed.ts
```diff
@@ -0,0 +1,56 @@
+import { Given, Then } from '@badeball/cypress-cucumber-preprocessor'
+import Page from '../../../../cypress_shared/pages/page'
+import BedspaceShowPage from '../../../../cypress_shared/pages/temporary-accommodation/manage/bedspaceShow'
+import LostBedNewPage from '../../../../cypress_shared/pages/temporary-accommodation/manage/lostBedNew'
+import newLostBedFactory from '../../../../server/testutils/factories/newLostBed'
+import lostBedFactory from '../../../../server/testutils/factories/lostBed'
+import LostBedShowPage from '../../../../cypress_shared/pages/temporary-accommodation/manage/lostBedShow'
+
+Given('I mark a bedspace as void', () => {
```
bryangaledxw: Sorry, realised when I was changing E2E tests myself I was talking nonsense about this! Think you had this right the first time as this doesn't actually mark the bedspace as void, it just goes to the page to do that. There's a rough convention of "I'm X-ing" meaning "I going to the page to do X but not actually doing X yet"

bryangaledxw (approved):
Looks good, and thanks for the changes! A few more comments, but mostly at the level of take-it-or-leave-it. Your call whether my suggestions would make the code better or not. There's still the stray change to integration tests in the E2E commit, and that commit message that references the type guard util, but I think this is basically ready to go in. Good stuff!

server/services/bookingService.ts
patrickjfl: ah yeah, true

cypress_shared/pages/temporary-accommodation/manage/bedspaceShow.ts
patrickjfl: yeah, exactly

e2e/tests/stepDefinitions/manage/lostBed.ts
patrickjfl: aha ok! have reverted that one (but not the second, for the reasons you give)

PR #186, opened 2023-02-27
Ensure bookings departures are never before arrivals

bryangaledxw:
We update our factories so that departures cannot be before arrivals. This fixes some random failures in our E2E tests, where the API validates that departures are not before arrivals


bryangaledxw:
[Fix random E2E test failure around bookings](https://trello.com/c/6QOqErkY/857-fix-random-e2e-test-failure-around-bookings)

tahb (approved):
Looks good

PR #187, opened 2023-02-28
Document what a bedspace is and how it came to be

tahb:
# Context

# Changes in this PR

As part of Pat's onboarding we discussed the value of documenting what a bedspace is and how that fits into the existing data modelling of rooms and beds within the API.

# Release checklist

confirm:

## Pre merge checklist

- [x] Have all changes to any dependencies been deployed to both `preprod` and
- [x] Has a required data migration been included in this change or been done in
    advance?

## Post merge checklist

considering it done.

fix it as soon as possible.

### DIFF
--- a/doc/architecture/decisions/0008-concept-of-bedspaces.md
+++ b/doc/architecture/decisions/0008-concept-of-bedspaces.md
@@ -1,0 +1,68 @@
+# 8. concept-of-bedspaces
+
+Date: 2023-02-28
+
+## Status
+
+Accepted
+
+## Context
+
+CAS1 and CAS3 share an API which includes the modelling for premises, rooms and
+beds. This modelling was initially set out by CAS1 and CAS3 have been working to
+extend in order to meet the needs of Temporary Accommodation (CAS3).
+
+The relationships are straightforward. A premises can have many rooms and a room
+can have many beds.
+
+For CAS1 they will often have the scenario of multiple beds belonging to a
+single room.
+
+For CAS3 we only expect one bed per room for the foreseeable future, however we
+cannot guarantee that this will always remain the case.
+
+At the time of writing an individual bed record doesn't itself contain any data.
+This too may change in future.
+
+Rooms serve to group related beds and to encapsulate shared attributes such as
+physical characteristics of the space. These are used in the matching journey
+where the risks and needs of person leaving prison are matched up to a suitable
+premises and room.
+
+Both teams have the need to help users create bookings for physical beds and to
+report on occupancy levels based on bed, rather than room. This points us
+towards using the existing bed model as the smallest unit of data for both
+services.
+
+Given the CAS3 team have 1 bed per room, we wanted to tailor the user experience
+to avoid unnecessary clicks. Instead of asking the user to navigate from
+premises through to rooms and on to the bed we can instead present all beds in a
+premises alongside their room characteristics. This removes the concept of
+'rooms' from our user.
+
+Finally, our users were more familiar with the language of 'bedspace' over
+'bed'.
+
+## Decision
+
+Temporary Accommodation will introduce the concept of a "bedspace" to users.
+
+## Consequences
+
+- Should we need to support rooms with multiple beds in future this should be
+  easier to introduce given the presence of the room model in the API
+- Viewing a bedspace will require the API to load both the room and the bedspace
+  which is less efficient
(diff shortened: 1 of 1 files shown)


doc/architecture/decisions/0008-concept-of-bedspaces.md
```diff
@@ -0,0 +1,62 @@
...
+
+Accepted
+
+## Context
+
+CAS1 and CAS3 share an API which includes the modelling for premises, rooms and
+beds. This modelling was initially set out by CAS1 and CAS3 have been working to
+extend in order to meet the needs of Temporary Accommodation (CAS3).
+
+The relationships are straight forward. A premises can have many rooms and a
```
bryangaledxw: "straight forward" -> "straightforward"

doc/architecture/decisions/0008-concept-of-bedspaces.md
```diff
@@ -0,0 +1,62 @@
...
+room can have many beds.
+
+For CAS1 they will often have the scenario of multiple beds belonging to a single
+room.
+
+For CAS3 we only expect one bed per room for the foreseeable future however
```
bryangaledxw: Think this would read better with "foreseeable future however" -> "foreseeable future, however we"

doc/architecture/decisions/0008-concept-of-bedspaces.md
```diff
@@ -0,0 +1,62 @@
...
+in a premises alongside their room characteristics. This removes the concept of
+'rooms' from our user.
+
+Finally, our users were more familiar with the language of 'bedspace' over 'bed'.
+
+## Decision
+
+Temporary Accommodation will introduce the concept of a "bedspace" to users.
+
+## Consequences
```
bryangaledxw, line 50: Suggest start each of these with a capital letter

doc/architecture/decisions/0008-concept-of-bedspaces.md
```diff
@@ -0,0 +1,62 @@
...
+The relationships are straight forward. A premises can have many rooms and a
+room can have many beds.
+
+For CAS1 they will often have the scenario of multiple beds belonging to a single
+room.
+
+For CAS3 we only expect one bed per room for the foreseeable future however
+cannot guarantee that this will always remain the case.
+
+At the time of writing an individual bed record doesn't itself contain any data,
```
bryangaledxw: I feel like this comma should either be a full stop or a dash

bryangaledxw (approved):
Makes sense and reads well, just some minor formatting feedback. 

And I wonder if we need one more point under consequences about how this works in practice in the codebase - i.e., work with rooms and ignore beds, unless we're forced by the API to use a bedId, in which case we do `room.beds[0].id`

doc/architecture/decisions/0008-concept-of-bedspaces.md
patrickjfl, line 50: It's maybe a bit fluffy as a consequence but one that I think is worth considering/documenting is that this decision probably causes a divergence between the concepts used by non-technical vs technical team members to think about and talk about the service. As in, technical team members kind of have to think about premises -> room -> bed -> booking, whereas non-technical people don't think about / talk about rooms, or maybe even know what they are?

Clearly this is a general issue with two services sharing an API, and also comes up in the case of Voids, which in our team are called voids, but in the code are called lostBeds right up until they are rendered in nunjucks.

patrickjfl (approved):
LGTM with one suggestion, thanks Tom.

doc/architecture/decisions/0008-concept-of-bedspaces.md
tahb: At some point my brain seems to have formed a habit of following [the GDS design system for bulleted lists](https://design-system.service.gov.uk/styles/typography/#bulleted-lists)! There doesn't seem to be any style guide for markdown but so I think it makes sense to do as you say 👍

tahb: I think that's fair. There is a cognitive burden for us to translate bedspaces into what they really represent and then back in a language the rest of the team can relate to. As you say that's also true of voids/lost beds, perhaps also worth an ADR? 

I'll add another point though, though the overhead could be implied there's no harm in making it explicit.

PR #197, opened 2023-03-02
Prevent double click on form buttons

bryangaledxw:
We add the preventDoubleClick attribute on form submission buttons across the service. While this cannot eliminate double-POSTing entirely, it can reduce its prevalence. We make an exception for the bookings report form button, as this does not take the user to a new page, and preventing the button here from being clicked more than once may be counter intuitive

### DIFF
--- a/server/views/applications/new.njk
+++ b/server/views/applications/new.njk
@@ -37,9 +37,10 @@
         }}
 
         {{ govukButton({
-                    name: 'submit',
-                    text: "Save and continue"
-                }) }}
+          name: 'submit',
+          text: "Save and continue",
+          preventDoubleClick: true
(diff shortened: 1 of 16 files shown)


bryangaledxw:
[Prevent double form submissions](https://trello.com/c/R9GvEJ9K/891-prevent-double-form-submissions)

patrickjfl (approved):
LGTM. So on the bookings report form, the user will download the report once for each click?

bryangaledxw:
@patrickjfl Trying it out, if I double click, I get the report once. If I click, wait for the report to download, and then click again, it'll download again. Which I feel like is the least surprising thing to do

PR #198, opened 2023-03-02
Propagate errors with empty 'invalid-params'

bryangaledxw:
We previously attempted to handle any API error with an 'invalid-params' array present. However, if this were empty, we would swallow the original API error, then throw a new error when attempting to render the empty errors array. This gave us the worst of both worlds - it didn't present a friendly error to the user, and it obscured the underlying error to developers.

We now check that the array is non-empty, and if it is empty, we simply allow the error to propagate.

### DIFF
--- a/server/utils/validation.test.ts
+++ b/server/utils/validation.test.ts
@@ -125,6 +125,23 @@ describe('catchValidationErrorOrPropogate', () => {
     expect(response.redirect).toHaveBeenCalledWith('some/url')
   })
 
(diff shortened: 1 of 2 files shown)


bryangaledxw:
[Propagate errors that contain an empty invalid-params array](https://trello.com/c/qko9UASN/890-propagate-errors-that-contain-an-empty-invalid-params-array)

tahb (approved):
LGTM

PR #196, opened 2023-03-02
Add more detail around the set up steps

tahb:
# Changes in this PR

This started out with the intention of documenting testing steps based on Pat's feedback. From working my way through the set up I felt there was enough value to add and get feedback whilst I continue looking at the testing steps.

# Release checklist

As part of our continuous deployment strategy we must ensure that this work is
ready to be released at any point. Before merging to `main` we must first
confirm:

## Pre merge checklist

- [ ] Have all changes to any dependencies been deployed to both `preprod` and
    `production` in a backwards compatible way? (This includes our API,
    infrastructure, third-party integrations and any secrets or environment variables)
- [ ] Has a required data migration been included in this change or been done in
    advance?

## Post merge checklist

Once we've merged we now need to release this through to production before
considering it done.

[Find the build-and-deploy job in CircleCI](https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-temporary-accommodation-ui)
and work through the following steps:

- [ ] Has the product manager asked to provide approval for this feature?
  - [ ] Has the product manager approved?
- [ ] Manually approve release to preprod
- [ ] Verify change released to preprod
- [ ] Manually approve release to prod
- [ ] Verify change released to production

Should a release fail at any step, you as the author should now lead the work to
fix it as soon as possible.

PR #199, opened 2023-03-02
Add more detail around our testing steps

tahb:
# Changes in this PR

Based upon https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/196 so please help to merge that first.

This change details the testing steps to help explain the differences of the numerous options.

As I said in the commit, the `test:e2e` commands don't work for me and I'm not sure what their original intent was, so I've left them out.

# Release checklist

confirm:

## Pre merge checklist

    advance?

## Post merge checklist

considering it done.

fix it as soon as possible.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,8 +22,6 @@
     "lint:fix": "eslint . --cache --max-warnings 0 --fix && node_modules/.bin/shellcheck ./script/**/*[^utils] ./script/utils/**",
     "typecheck": "tsc && tsc -p integration_tests",
     "test": "jest",
-    "test:e2e": "cypress run -C cypress.config.e2e.ts",
-    "test:e2e:ui": "set -a && source e2e.env && set +a && cypress open --e2e --browser electron -C cypress.config.e2e.ts --config baseUrl=https://approved-premises-dev.hmpps.service.justice.gov.uk",
     "test:integration": "npm run start-test-wiremock && start-server-and-test start-feature http://localhost:3007/ping int-test",
     "test:integration:ui": "npm run start-test-wiremock && start-server-and-test start-feature:dev http://localhost:3007/ping int-test-ui",
     "security_audit": "npx audit-ci --config audit-ci.json",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
     "test": "jest",
     "test:e2e": "cypress run -C cypress.config.e2e.ts",
     "test:e2e:ui": "set -a && source e2e.env && set +a && cypress open --e2e --browser electron -C cypress.config.e2e.ts --config baseUrl=https://approved-premises-dev.hmpps.service.justice.gov.uk",
-    "test:integration": "start-server-and-test start-feature http://localhost:3007/ping int-test",
+    "test:integration": "npm run start-test-wiremock && start-server-and-test start-feature http://localhost:3007/ping int-test",
```
bryangaledxw, line 25: I nearly did this tidy up myself, then backed out given that `test:integration` is run in `script/test`, and I wasn't sure what would happen if that script spun up wiremock via `docker-compose`, spun wiremock up again via `test:integration`, then only cleaned up the latter. Happy that this is safe?

bryangaledxw (commented):
I feel like this might be best practice that I'm not aware of, and this isn't a strongly held objection, but I'm not sure of the need to keep the E2E environment setup out of git? I feel like if it needs to contain sensitive data, that's an issue in itself?

Is this correct? - "I found that the .e2e.env file is git ignored but also required by the local_e2e script" - I don't have `.e2e.env` anywhere and it looks like `local_e2e` only references `e2e.local.env` right now

And yep, I never use `npm run test:e2e`. I think it's assuming everything is already up and running via ap tools?

package.json
tahb: I think it's ok? It looks to me like both `test:integration` and `script/test` attempt to start the backing services via docker-compose but only `script/test` cleans up after itself. 

If `script/test` runs after a lone request to do `test:integration` then it should find existing docker containers and continue as normal. I've given this a test and it seems to be happy!

tahb:
As we're happy to drop the following tasks - which were the remaining reference to e2e.env - I think we can go back to e2e.local:

```
"test:e2e"
"test:e2e:ui"
```

I agree that e2e.env should include safe values that we can commit to git. Looks like we are inheriting the decision to include that file in the gitignore. Feels risky to unignore that file now just incase anyone has a file they might then commit.

bryangaledxw (approved):
LGTM

PR #202, opened 2023-03-06
Fix filterBlanksFromData

bryangaledxw:
We have a `filterBlanksFromData` method which strips out any empty variables that are passed to the API. However, this also strips out any literal `false` values, which we do want to send.

This updates the `filterBlanksFromData` helper to check if the item is a literal boolean before checking it is ‘falsy’ (i.e. null, undefined or a blank string). This ensures any literal `false` values are sent to the API as expected.

This is based on [Approved Premises' similar PR](https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/594), though we do not make similar changes to the Apply integration tests, as TA is still relying on the API extracting the fields they test, rather than the UI sending the,

### DIFF
--- a/server/data/restClient.test.ts
+++ b/server/data/restClient.test.ts
@@ -2,8 +2,8 @@ import { createMock } from '@golevelup/ts-jest'
 import { Response } from 'express'
 import nock from 'nock'
 import type { ApiConfig } from '../config'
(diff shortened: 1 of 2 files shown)


patrickjfl (approved):
Nice 👍

PR #207, opened 2023-03-07
API model updates

pezholio:


bryangaledxw (approved):
Looks good. What might be worth doing though is `import type { TemporaryAccommodationLostBed as LostBed }`, to cut down on some of the verbosity. We're doing that with TemporaryAccommodationPremises

server/testutils/factories/lostBed.ts
```diff
@@ -1,16 +1,18 @@
...
-export default Factory.define<LostBed>(() => ({
+export default Factory.define<TemporaryAccommodationLostBed>(() => ({
   id: faker.datatype.uuid(),
   notes: faker.lorem.sentence(),
   startDate: DateFormats.dateObjToIsoDate(faker.date.soon()),
   endDate: DateFormats.dateObjToIsoDate(faker.date.future()),
   numberOfBeds: faker.datatype.number({ max: 10 }),
   referenceNumber: faker.datatype.uuid(),
   reason: referenceDataFactory.lostBedReasons().build(),
+  status: faker.helpers.arrayElement(['active', 'cancelled'] as const),
```
bryangaledxw: Also, is the `as const` needed? It seemed to compile fine for me without that

patrickjfl:
> Looks good. What might be worth doing though is `import type { TemporaryAccommodationLostBed as LostBed }`, to cut down on some of the verbosity. We're doing that with TemporaryAccommodationPremises

Yeah, makes sense.

server/testutils/factories/lostBed.ts
patrickjfl: no, you're right - removed

PR #213, opened 2023-03-09
Update redis dependency

bryangaledxw:
Recreation of #188. The automatic redis update needed some manual intervention. Recreating as a new PR so Renovate doesn't keep overwriting manual changes


bryangaledxw:
[Update redis dependency](https://trello.com/c/zhrUDd6C/889-update-redis-dependency)

package.json
```diff
@@ -177,7 +177,6 @@
     "@types/bunyan-format": "^0.2.5",
     "@types/compression": "^1.7.2",
     "@types/connect-flash": "0.0.37",
-    "@types/connect-redis": "^0.0.20",
```
tahb, line 178: Was this part of the manual intervention we had to do? Do we not need this?

tahb (approved):
LGTM. One query. 

I fell into a bit of a rabbit hole trying to observe redis continuing to work. I turned off the redis backing service and it certainly started breaking so seems like it's still connecting and there's no fallback.

package.json
pezholio: The types are now included in the library, so we don't need this 👍

PR #217, opened 2023-03-14
Users can edit and cancel a void booking

patrickjfl:
# Context

Allows users to [edit and cancel a void booking](https://trello.com/c/AYpNNB2q/757-users-can-edit-and-cancel-a-void-booking). For now the changes remain feature flagged and disabled outside of dev, test and local environments.

End to end testing is not included in this PR but will follow shortly.

# Changes in this PR

## Screenshots of UI changes

View void booking screen, with edit/cancel actions:
Screenshot 2023-03-14 at 11 08 36

Edit void booking screen:
Screenshot 2023-03-14 at 11 09 27

View void booking screen, with update success message:
Screenshot 2023-03-14 at 11 09 41

Cancel void booking screen:
Screenshot 2023-03-14 at 11 10 04

View void booking screen, with cancellation success message and cancelled status:
Screenshot 2023-03-14 at 11 10 10

# Release checklist

confirm:

## Pre merge checklist

    advance?

## Post merge checklist

considering it done.

fix it as soon as possible.

### DIFF
--- a/integration_tests/tests/temporary-accommodation/manage/lostBed.cy.ts
+++ b/integration_tests/tests/temporary-accommodation/manage/lostBed.cy.ts
@@ -1,3 +1,5 @@
+import newLostBedCancellationFactory from '../../../../server/testutils/factories/newLostBedCancellation'
+import updateLostBedFactory from '../../../../server/testutils/factories/updateLostBed'
 import Page from '../../../../cypress_shared/pages/page'
 import BedspaceShowPage from '../../../../cypress_shared/pages/temporary-accommodation/manage/bedspaceShow'
 import LostBedNewPage from '../../../../cypress_shared/pages/temporary-accommodation/manage/lostBedNew'
@@ -7,6 +9,8 @@ import lostBedFactory from '../../../../server/testutils/factories/lostBed'
 import newLostBedFactory from '../../../../server/testutils/factories/newLostBed'
 import premisesFactory from '../../../../server/testutils/factories/premises'
 import roomFactory from '../../../../server/testutils/factories/room'
+import LostBedEditPage from '../../../../cypress_shared/pages/temporary-accommodation/manage/lostBedEdit'
+import LostBedCancelPage from '../../../../cypress_shared/pages/temporary-accommodation/manage/lostBedCancel'
 
 context('Lost bed', () => {
   beforeEach(() => {
@@ -117,7 +121,7 @@ context('Lost bed', () => {
     lostBedShowPage.shouldShowBanner('Void created')
   })
 
-  it('shows errors when the API returns an error', () => {
+  it('shows create errors when the API returns an error', () => {
     // Given I am signed in
     cy.signIn()
 
@@ -143,7 +147,7 @@ context('Lost bed', () => {
     page.shouldShowErrorMessagesForFields(['startDate', 'endDate'])
   })
 
-  it('shows errors when the API returns a 409 conflict error', () => {
+  it('shows create errors when the API returns a 409 conflict error', () => {
     // Given I am signed in
     cy.signIn()
 
@@ -197,14 +201,34 @@ context('Lost bed', () => {
     Page.verifyOnPage(BedspaceShowPage, premises, room)
   })
 
-  it('shows a single lost bed', () => {
+  it('shows a single active lost bed', () => {
     // Given I am signed in
     cy.signIn()
 
-    // And there is a premises, a room, and a lost bed in the database
+    // And there is a premises, a room, and an active lost bed in the database
     const premises = premisesFactory.build()
     const room = roomFactory.build()
-    const lostBed = lostBedFactory.build()
+    const lostBed = lostBedFactory.active().build()
+
+    cy.task('stubSinglePremises', premises)
+    cy.task('stubSingleRoom', { premisesId: premises.id, room })
+    cy.task('stubSingleLostBed', { premisesId: premises.id, lostBed })
+
+    // When I visit the show lost bed page
+    const page = LostBedShowPage.visit(premises, room, lostBed)
+
+    // Then I should see the booking details
+    page.shouldShowLostBedDetails()
+  })
+
+  it('shows a single cancelled lost bed', () => {
+    // Given I am signed in
+    cy.signIn()
+
+    // And there is a premises, a room, and a cancelled lost bed in the database
+    const premises = premisesFactory.build()
+    const room = roomFactory.build()
+    const lostBed = lostBedFactory.build({ status: 'cancelled' })
 
     cy.task('stubSinglePremises', premises)
     cy.task('stubSingleRoom', { premisesId: premises.id, room })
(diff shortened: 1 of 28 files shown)


patrickjfl:
[Users can edit and cancel a void booking](https://trello.com/c/AYpNNB2q/757-users-can-edit-and-cancel-a-void-booking)

server/@types/shared/models/LostBedCancellation.ts
```diff
@@ -3,8 +3,7 @@
...
 export type LostBedCancellation = {
-    id: string;
-    createdAt: string;
-    notes?: string;
-};
-
+  id: string
+  createdAt: string
+  notes?: string
+}
```
bryangaledxw: Able to rerun `generate-types`, and squash the result into this commit? Just to keep `server/@types` as the unmodified output from that script

server/services/lostBedService.ts
```diff
@@ -29,6 +32,44 @@ export default class LostBedService {
...
+    lostBedId: string,
+    newLostBedCancellation: NewLostBedCancellation,
+  ): Promise<LostBedCancellation> {
+    const lostBedClient = this.lostBedClientFactory(callConfig)
+
+    const lostBedCancellation = await lostBedClient.cancel(premisesId, lostBedId, newLostBedCancellation)
+
+    return lostBedCancellation
+  }
+
```
bryangaledxw, line 72: `getUpdateLostBed` makes sense as the method is as much about the type conversion as anything else. `updateLostBed` and `cancelLostBed`, I'd be tempted to rename to just `update` and `cancel`

server/paths/api.ts
```diff
@@ -5,7 +5,7 @@ const singlePremisesPath = premisesPath.path(':premisesId')
 
 const lostBedsPath = singlePremisesPath.path('lost-beds')
 const singleLostBedPath = lostBedsPath.path(':lostBedId')
-const cancelLostBedPath = singleLostBedPath.path('cancellation')
+const cancelLostBedPath = singleLostBedPath.path('cancellations')
```
bryangaledxw: Think this belongs in the first commit

server/paths/temporary-accommodation/manage.ts
```diff
@@ -67,6 +67,9 @@ const paths = {
     new: lostBedsPath.path('new'),
     create: lostBedsPath,
     show: singleLostBedPath,
+    edit: singleLostBedPath.path('edit'),
+    update: singleLostBedPath,
+    cancellations: { new: singleLostBedPath.path('cancel'), create: singleLostBedPath.path('cancellations') },
```
bryangaledxw, line 72: I feel like you've probably considered this and decided against, but my instinct here would be to keep this constent with bookings, so:

```
cancellations: { new: lostBedCancellationsPath.path('new'), create: lostBedCancellationsPath },
```

server/views/temporary-accommodation/lost-beds/cancel.njk
```diff
@@ -0,0 +1,40 @@
...
+            },
+            fieldName: "notes"
+          },
+          fetchContext()
+        ) }}
+    {{ govukButton({
+            text: "Submit"
+        }) }}
+  </form>
+{% endblock %}
```
bryangaledxw: Let's end with a newline - this always looks a like a warning to me. And I'd prefer some vertical whitespace between the sections. Same for `edit.njk`

server/views/temporary-accommodation/lost-beds/cancel.njk
```diff
@@ -0,0 +1,40 @@
...
+{% from "../components/ta-textarea/macro.njk" import taTextarea %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% extends "../../partials/layout.njk" %}
+{% set pageTitle = applicationName + " - Cancel void booking" %}
+{% set mainClasses = "app-container govuk-body" %}
+{% block content %}
+  {{ breadCrumb('Cancel a void booking', [
+    {title: 'List of bedspaces', href: paths.premises.show({ premisesId: premises.id })},
+    {title: 'View bedspace', href: paths.premises.bedspaces.show({ premisesId: premises.id, roomId: room.id })}
+  ]) }}
```
bryangaledxw, line 19: I don't know if I can justify my intuition here, but I feel like this is a `govukBackLink`, rather than `breadCrumb`. Here and for `edit.njk` My mental model is that there's a tree of pages, then things like this and booking state changes are these odd little ephemeral spindles off of that. Like, they're definitely not permanent parts of the tree, they're only here for particular states.

I think maybe that's my rule - "permanent pages", pages that might make sense to bookmark, breadcrumbs. Otherwise, it wants a backlink

server/controllers/temporary-accommodation/manage/lostBedsController.ts
```diff
@@ -81,8 +85,121 @@ export default class LostBedsController {
...
+      } catch (err) {
+        catchValidationErrorOrPropogate(
+          req,
+          res,
+          err,
+          paths.lostBeds.cancellations.new({ premisesId, roomId, lostBedId }),
+        )
+      }
+    }
+  }
```
bryangaledxw, line 199: Can go either way, since I don't think this is going to expand to a bunch of different actions like bookings, but consider if it's worth pulling these two methods into their own controller

server/controllers/temporary-accommodation/manage/lostBedsController.ts
```diff
@@ -81,8 +85,121 @@ export default class LostBedsController {
         premises,
         room,
         lostBed,
+        actions: lostBedActions(premisesId, roomId, lostBed),
         allStatuses,
       })
     }
   }
+
+  update(): RequestHandler {
```
bryangaledxw, line 94: Would swap the order of `update` and `edit` here and in the test, just to reflect user journey order

server/components/lostBedInfo.ts
```diff
@@ -26,7 +26,10 @@ export default (lostBed: LostBed): SummaryList['rows'] => {
 
   rows.push({
     key: textValue('Notes'),
-    value: htmlValue(formatLines(lostBed.notes)),
+    value:
+      lostBed.status === 'active'
+        ? htmlValue(formatLines(lostBed.notes))
+        : htmlValue(formatLines(lostBed.cancellation.notes)),
```
bryangaledxw, line 32: Suggest reflect this change in `lostBedInfo.test.ts`

server/routes/temporary-accommodation/manage.ts
```diff
@@ -200,6 +200,35 @@ export default function routes(controllers: Controllers, services: Services, rou
...
+    })
+    post(paths.lostBeds.cancellations.create.pattern, lostBedsController.createCancellation(), {
+      redirectAuditEventSpecs: [
+        {
+          path: paths.lostBeds.cancellations.new.pattern,
+          auditEvent: 'CREATE_LOST_BED_FAILURE',
+        },
+        {
+          path: paths.lostBeds.show.pattern,
+          auditEvent: 'CREATE_LOST_BED_SUCCESS',
```
bryangaledxw: These should be unique

integration_tests/tests/temporary-accommodation/manage/lostBed.cy.ts
```diff
@@ -197,14 +201,34 @@ context('Lost bed', () => {
...
+  it('shows a single active lost bed', () => {
     // Given I am signed in
     cy.signIn()
 
-    // And there is a premises, a room, and a lost bed in the database
+    // And there is a premises, a room, and an active lost bed in the database
     const premises = premisesFactory.build()
     const room = roomFactory.build()
-    const lostBed = lostBedFactory.build()
+    const lostBed = lostBedFactory.build({ status: 'active' })
```
bryangaledxw: Definitely not a blocker, but consider expanding the factory so we can just do `lostBedFactory.active().build()` here

integration_tests/tests/temporary-accommodation/manage/lostBed.cy.ts
```diff
@@ -244,4 +268,153 @@ context('Lost bed', () => {
...
+    page.completeForm(cancelLostBed)
+
+    // Then the lost bed should have been cancelled in the API
+    cy.task('verifyLostBedCancel', { premisesId: premises.id, lostBedId: lostBed.id }).then(requests => {
+      expect(requests).to.have.length(1)
+      const requestBody = JSON.parse(requests[0].body)
+
+      expect(requestBody.notes).equal(cancelLostBed.notes)
+    })
+  })
```
bryangaledxw, line 466: May be worth adding a couple of tests to check that going back from cancel/edit go to the right place

server/services/lostBedService.test.ts
```diff
@@ -10,6 +10,9 @@ import ReferenceDataClient from '../data/referenceDataClient'
 import lostBedFactory from '../testutils/factories/lostBed'
```
bryangaledxw, line 10: Check the message in this commit - think that comma wants to be a full stop

bryangaledxw (approved):
Good stuff! Some feedback, but only must-fix is the audit event names.

I was trying a failing to decide if I'd put the new controller method in their own controllers. It feel like we should do the same thing with both edit and cancel. But then edit feels like it belongs in the existing controller, but cancel maybe feels more like it belongs on its own, at least to be consistent with bookings. Anyway, I'm good for it to stay as is if you're happy with it.

I think don't worry about it for this PR, but for something like this, I'd maybe have added update (and testing that) as one block of commits, then added cancel (and its testing), so we're adding a new thing and its testing as close together as we can

server/paths/api.ts
patrickjfl: ah thanks :)

server/paths/temporary-accommodation/manage.ts
patrickjfl: Yeah, I see your point, I think it just feels kind of misleading to/confusing for users if the URL implies that they can make multiple cancellations of a single void.

server/views/temporary-accommodation/lost-beds/cancel.njk
patrickjfl: ah yeah, nunjucks again. How do you format yours?

server/routes/temporary-accommodation/manage.ts
patrickjfl: oh god, good spot. I think that got mixed up in rebasing

integration_tests/tests/temporary-accommodation/manage/lostBed.cy.ts
patrickjfl: huh, I already did that but just forgot about it! Will amend

integration_tests/tests/temporary-accommodation/manage/lostBed.cy.ts
patrickjfl: good call

server/views/temporary-accommodation/lost-beds/cancel.njk
bryangaledxw: I feel like at a minimum, after the `{% from ... %}`s, after the `{% extends ... %}`, and before any `{% block %}`s, and between the "header" bits of the page and anything else.

I did go down the rabbit hole of automatic Nunjucks linting on my last dxw project, but couldn't find anything that wasn't bad in some way or another!

PR #218, opened 2023-03-14
Restore test:e2e command

bryangaledxw:
This is used as part of our CI process, and is referenced in config.yaml. We restore this after being inadvertently deleted in a previous commit


tahb (approved):
We dicussed renaming this to something like `test:ci:e2e` to better convey the usage of this one away from devs using it locally. Approving in advance 👍

bryangaledxw:
@tahb Have gone with `test:e2e:ci`, to keep `test:<type>:<context>` format

PR #229, opened 2023-03-20
Ensure booking confirmation screen handles all classes of CRN error

bryangaledxw:
We update our Content Security Policy to allow the HMPPS Auth URL as a possible form action target.

Currently, if a 403 error occurs on a GET request, this will be captured by the error handling setup in errorHandler.ts, and the user will be redirected to the sign out URL, which then redirects to the HMPSS Auth URL. However, if a 403 error occurs on a POST request, this second redirect may not occur, and the user may, depending on their choice of browser, be frozen on the form page they just submitted.

Due to CSP implementation details that vary between browsers, adding the HMPPS Auth URL to our form action targets allows this second redirect to occur as expected.

For further details, see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/form-action

### DIFF
--- a/server/middleware/setUpWebSecurity.ts
+++ b/server/middleware/setUpWebSecurity.ts
@@ -1,6 +1,6 @@
+import crypto from 'crypto'
 import express, { NextFunction, Request, Response, Router } from 'express'
 import helmet from 'helmet'
-import crypto from 'crypto'
 
 export default function setUpWebSecurity(): Router {
   const router = express.Router()
--- a/server/middleware/setUpWebSecurity.ts
+++ b/server/middleware/setUpWebSecurity.ts
@@ -1,6 +1,7 @@
 import crypto from 'crypto'
 import express, { NextFunction, Request, Response, Router } from 'express'
 import helmet from 'helmet'
+import config from '../config'
 
 export default function setUpWebSecurity(): Router {
   const router = express.Router()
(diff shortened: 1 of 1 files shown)


tahb:
Nice work. This seems reasonable to me. 

hmpps-auth is a trusted domain and so we should be able to redirect there safely. Any of our form data made visible there shouldn't be an issue. If I understand this correctly then even if hmpps-auth added another redirect to somewhere other than us, such as evil.com, then the redirect wouldn't work.

There appears to be some precedent for this in 3 other projects. This makes it feel like a good candidate to go back into the the template:

- https://github.com/ministryofjustice/book-a-prison-visit-staff-ui/blob/7c9997ade13f599e3c8cac0d92ac176a097edff8/server/middleware/setUpWebSecurity.ts#L20
- https://github.com/ministryofjustice/make-recall-decision-ui/blob/e09b64f52eb908f2a9f43ca8e0df13c4202e7c23/server/middleware/setUpWebSecurity.ts#L31
- https://github.com/ministryofjustice/hmpps-incentives-ui/blob/b601aea8900735a78b003c623083d35f4b8b19e9/server/middleware/setUpWebSecurity.ts#L22

tahb (approved):

bryangaledxw:
@tahb Ah thanks, didn't think to check what other projects are doing here!

PR #231, opened 2023-03-20
Fix local authority name

bryangaledxw:
Fix a local authority name in our JSON data, which could lead to E2E test failures


patrickjfl (approved):
Enthusiastic approval from this Herefordian 👍

PR #258, opened 2023-03-29
E2E tests read reference data from page

bryangaledxw:
We want to update our premises editing page to only display PDUs that are relevant to the current user.

By default, we would expect this to break our E2E tests, as when creating/editing a premises, they first construct a Premises object using reference data known by the tests, and then use this to guide creating/editing the premises. More often than not, we'd expect the PDU chosen by the E2E tests to not exist in the PDU dropdown shown to the test user, causing an error when the E2E test tries to select a PDU that isn't there.

To fix this, and to make our tests more robust against future changes to PDUs, rather than relying on a prior, known set of PDUs, we read the available PDUs from the page. We also take this as an opportunity to make a similar change for other reference data we need to know when creating/editing a premises.

### DIFF
--- a/server/testutils/factories/premises.ts
+++ b/server/testutils/factories/premises.ts
@@ -2,7 +2,11 @@ import { faker } from '@faker-js/faker/locale/en_GB'
 import { Factory } from 'fishery'
 
 import type { ApArea, TemporaryAccommodationPremises as Premises } from '@approved-premises/api'
+import { ReferenceData } from '../../@types/ui'
 import { unique } from '../../utils/utils'
+import characteristicFactory from './characteristic'
+import localAuthorityFactory from './localAuthority'
+import probationRegionFactory from './probationRegion'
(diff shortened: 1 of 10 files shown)


bryangaledxw:

server/testutils/factories/premises.ts
```diff
@@ -17,6 +21,32 @@ class PremisesFactory extends Factory<Premises> {
       status: 'archived',
     })
   }
+
+  /* istanbul ignore next */
+  forEnvironment(
```
patrickjfl, line 26: Should this be forEnvironment or fromEnvironment? I think either is ok but just to point out the inconsistency between the method name and the commit description

patrickjfl (approved):
Nice job 👍

server/testutils/factories/premises.ts
bryangaledxw: Ah thanks. Going with `forEnvironment`

PR #259, opened 2023-03-30
Update bedspace search and premises editing to only present PDUs associated with users region

bryangaledxw:
# Changes in this PR

This PR updates the UI to retrieve the list of PDUs from an API endpoint, rather than using a UI-side JSON file. This allows us to send the API a query parameter that the API will use to send us only the PDUs for our region.

This PR is ahead of the API, and makes some assumptions about the API and the shared types. The assumed shared types are in the "Update shared types" commit, and should be replaced with the 'real' shared types before merging

## API assumptions

I've generally assumed this will follow the lead of the bedspace search work, and move towards using "probation delivery unit", rather than the abbreviated "PDU".

- We'll have a new type `ProbationDeliveryUnit`, with fields `name` and `id`
- `pdu` on `TemporaryAccommodationPremises` will continue to give the PDU *name*, for backwards compatibility
- `TemporaryAccommodationPremises` will have a new field `probationDeliveryUnit`, which will be a `ProbationDeliveryUnit` object, similar to `probationRegion` and `localAuthorityArea`
- The PDUs for a region will be available at `/reference-data/probation-delivery-units?probationRegionId=<< probation region ID >>`, and this will return an array of `ProbationDeliveryUnit`s

On `NewPremises` and `UpdatePremises`:
- We deprecate `pdu`, and use a new `probationDeliveryUnitId` field
- If `pdu` is present, it's coming from the old UI and should be interpreted as a PDU *name*. If `probationDeliveryUnitId` is present it's coming from the new UI and should be interpreted as a PDU *ID*.
- **If both are missing, the "empty" error goes against the `probationDeliveryUnitId` field**
- A frontend update to go in before the API change to handle `probationDeliveryUnitId` errors

Note that the PDUs initially seeded in the dev environment don't match those the UI is aware of from [`pdu.json`](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/main/server/data/pdus.json). E.g., in our JSON we have "Essex South". In the [dev seed](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/src/main/resources/db/migration/local%2Bdev/R__2_create_premises.sql) we have "Essex South (Basildon, Brentwood, Rochford and Castlepoint, Southend-on-Sea)".

# Release checklist

confirm:

## Pre merge checklist

    advance?

## Post merge checklist

considering it done.

fix it as soon as possible.

### DIFF
--- a/cypress_shared/pages/temporary-accommodation/manage/premisesEdit.ts
+++ b/cypress_shared/pages/temporary-accommodation/manage/premisesEdit.ts
@@ -46,7 +46,7 @@ export default class PremisesEditPage extends PremisesEditablePage {
       .contains('What is the PDU?')
       .siblings('select')
       .children('option')
-      .contains(exact(this.premises.pdu))
+      .contains(exact(this.premises.probationDeliveryUnit.name))
       .should('be.selected')
 
     cy.get('legend')
--- a/cypress_shared/pages/temporary-accommodation/manage/premisesEdit.ts
+++ b/cypress_shared/pages/temporary-accommodation/manage/premisesEdit.ts
@@ -85,7 +85,7 @@ export default class PremisesEditPage extends PremisesEditablePage {
 
     this.getSelectInputByIdAndSelectAnEntry('localAuthorityAreaId', '')
     this.getSelectInputByIdAndSelectAnEntry('probationRegionId', '')
-    this.getSelectInputByIdAndSelectAnEntry('pdu', '')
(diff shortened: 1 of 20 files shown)


bryangaledxw:
[Update bedspace search to only present PDUs associated with user's region](https://trello.com/c/AXdBg54M/972-update-bedspace-search-to-only-present-pdus-associated-with-users-region)

bryangaledxw:

PR #261, opened 2023-03-30
Add factories to an index

bryangaledxw:
Modelled after a [similar PR in AP](https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/666), we add our factories classes to an index to help keep naming consistent and make them easier to use in VS code


patrickjfl (approved):
Nice 👍

PR #281, opened 2023-04-12
API model updates

pezholio:

### DIFF
--- a/integration_tests/tests/temporary-accommodation/manage/bookingSearch.cy.ts
+++ b/integration_tests/tests/temporary-accommodation/manage/bookingSearch.cy.ts
@@ -1,3 +1,4 @@
+import type { BookingStatus } from '@approved-premises/api'
 import DashboardPage from '../../../../cypress_shared/pages/temporary-accommodation/dashboardPage'
 import BookingSearchPage from '../../../../cypress_shared/pages/temporary-accommodation/manage/bookingSearch'
 import setupTestUser from '../../../../cypress_shared/utils/setupTestUser'
@@ -37,7 +38,7 @@ context('Booking search', () => {
     // And there are bookings of all relevant status types in the database
     const bookings = bookingSearchResultsFactory.build()
 
-    const statuses = ['provisional', 'active', 'closed', 'confirmed']
+    const statuses: Array<BookingStatus> = ['provisional', 'arrived', 'departed', 'confirmed']
 
     statuses.forEach(status => {
       cy.task('stubFindBookings', { bookings, status })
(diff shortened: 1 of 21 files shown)


integration_tests/tests/temporary-accommodation/manage/bookingSearch.cy.ts
```diff
@@ -37,7 +37,7 @@ context('Booking search', () => {
     // And there are bookings of all relevant status types in the database
     const bookings = bookingSearchResultsFactory.build()
 
-    const statuses = ['provisional', 'active', 'closed', 'confirmed']
+    const statuses = ['provisional', 'arrived', 'departed', 'confirmed']
```
richpjames: I suggest typing this `Array<BookingUiStatuses>` so this can't contain items that aren't booking statuses

server/controllers/temporary-accommodation/manage/bookingSearchController.ts
```diff
@@ -1,20 +1,33 @@
...
-  index(status: BookingSearchStatus): RequestHandler {
+  index(status: BookingStatus): RequestHandler {
     return async (req: Request, res: Response) => {
       const callConfig = extractCallConfig(req)
 
       const bookingTableRows = await this.bookingSearchService.getTableRowsForFindBooking(callConfig, status)
 
+      let uiStatus: string
+
+      switch (status) {
```
richpjames: I'm probably missing something but it would be a lot easier if we could just use 'arrived' and 'departed' as the status instead of 'active' and 'closed', especially as this is what it will say in the table headings and side nav

integration_tests/tests/temporary-accommodation/manage/bookingSearch.cy.ts
patrickjfl: good shout

server/controllers/temporary-accommodation/manage/bookingSearchController.ts
patrickjfl: it would be a lot easier :) but for now we have two sets of terminology, one that comes from the API ('arrived', 'departed') and another that comes from our users ('active', 'closed'). So one way or another we do need to translate from one to the other in our UI (and in fact we do the same in generating the table headings and side nav, although maybe we could refactor that so it's in one place...)

richpjames: I thought as much. 
I think it would make most sense if the functions that did the translating were in the same place and the translations happened at the same time so that if we have to change them in future it's easy

richpjames (approved):

server/controllers/temporary-accommodation/manage/bookingSearchController.ts
patrickjfl: good shout, done

PR #292, opened 2023-04-17
Rename closed booking status to departed

patrickjfl:
# Changes in this PR

This PR renames the 'closed' UI booking status to departed, in advance of further anticipated changes to booking state as part of turnaround development.

It also includes some minor refactoring to remove an unnecessary UI type.

# Release checklist

confirm:

## Pre merge checklist

    advance?

## Post merge checklist

considering it done.

fix it as soon as possible.

### DIFF
--- a/server/routes/temporary-accommodation/manage.ts
+++ b/server/routes/temporary-accommodation/manage.ts
@@ -268,8 +268,8 @@ export default function routes(controllers: Controllers, services: Services, rou
   get(paths.bookings.search.active.index.pattern, bookingSearchController.index('arrived'), {
     auditEvent: 'VIEW_SEARCH_ACTIVE_BOOKINGS',
   })
-  get(paths.bookings.search.closed.index.pattern, bookingSearchController.index('departed'), {
-    auditEvent: 'VIEW_SEARCH_CLOSED_BOOKINGS',
+  get(paths.bookings.search.departed.index.pattern, bookingSearchController.index('departed'), {
+    auditEvent: 'VIEW_SEARCH_DEPARTED_BOOKINGS',
   })
   get(paths.bookings.search.confirmed.index.pattern, bookingSearchController.index('confirmed'), {
     auditEvent: 'VIEW_SEARCH_CONFIRMED_BOOKINGS',
--- a/server/utils/bookingSearchUtils.ts
+++ b/server/utils/bookingSearchUtils.ts
@@ -20,9 +20,9 @@ export function createSideNavArr(status: BookingSearchUiStatus): Array<SideNavOb
       active: status === 'active',
     },
     {
-      text: 'Closed',
-      href: paths.bookings.search.closed.index({}),
-      active: status === 'closed',
+      text: 'Departed',
+      href: paths.bookings.search.departed.index({}),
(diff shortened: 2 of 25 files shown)


server/routes/temporary-accommodation/manage.ts
```diff
@@ -268,8 +268,8 @@ export default function routes(controllers: Controllers, services: Services, rou
   get(paths.bookings.search.active.index.pattern, bookingSearchController.index('arrived'), {
     auditEvent: 'VIEW_SEARCH_ACTIVE_BOOKINGS',
   })
-  get(paths.bookings.search.closed.index.pattern, bookingSearchController.index('departed'), {
-    auditEvent: 'VIEW_SEARCH_CLOSED_BOOKINGS',
+  get(paths.bookings.search.departed.index.pattern, bookingSearchController.index('departed'), {
+    auditEvent: 'VIEW_SEARCH_DEPARTED_BOOKINGS',
```
bryangaledxw, line 272: Double check with @tahb on this one - are we ok with changing the vocab in our audit logs?

bryangaledxw (approved):
Looks good, but double check with Tom about changing audit vocab. I suspect it's fine while this is beta, but I reckon it's something where ideally we want to be to interpret historic audit data without having to account for terminology changing.

And to confirm my understanding, there's no API changes for this, right?

server/utils/bookingSearchUtils.ts
```diff
@@ -63,8 +63,6 @@ export function convertApiStatusToUiStatus(status: BookingSearchApiStatus): Book
   switch (status) {
```
bryangaledxw, line 65: Looking into function, do we need BookingSearchUIStatus? Looks like all we need is a function for turning an API Booking Status into a human readable string?

server/routes/temporary-accommodation/manage.ts
patrickjfl: will do

server/utils/bookingSearchUtils.ts
patrickjfl: Yeah arguably not. Good call, I'll refactor that

patrickjfl:
> Looks good, but double check with Tom about changing audit vocab. I suspect it's fine while this is beta, but I reckon it's something where ideally we want to be to interpret historic audit data without having to account for terminology changing.
> 
> And to confirm my understanding, there's no API changes for this, right?

Yes, that's right. My understanding is we're renaming 'closed' to 'departed' to make space for a new 'closed' category that we are introducing in both the API and the UI.

server/routes/temporary-accommodation/manage.ts
tahb: Think this is fine as the 'search bookings' isn't on in prod yet. 

If it were on prod and was for a significant event which tracked a user viewing personal data then we _could_ look at asking the audit team to update old event keys.

patrickjfl: Tom says he is happy with this given that the changes are yet to be released to prod.

patrickjfl: OK, thanks Tom

PR #295, opened 2023-04-17
E2E tests read bedspace attributes from page

bryangaledxw:
# Changes in this PR

- We update to UI E2E tests, so we no longer need to keep `characteristics.json` file synchronised with the API in order for the tests to pass. We instead derive this data from the bedspace creation and editing pages

# Release checklist

[Release process

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

- [ ] [Manually
  release to test
- [ ] [Manually
  release to preprod
- [ ] [Manually
  release to prod

### DIFF
--- a/cypress_shared/pages/temporary-accommodation/manage/bedspaceEditable.ts
+++ b/cypress_shared/pages/temporary-accommodation/manage/bedspaceEditable.ts
@@ -12,4 +12,8 @@ export default abstract class BedspaceEditablePage extends Page {
 
     this.clickSubmit()
   }
+
+  assignCharacteristics(alias: string): void {
+    this.getCheckboxItemsAsReferenceData('Does the bedspace have any of the following attributes?', alias)
+  }
 }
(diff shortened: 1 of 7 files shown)


cypress_shared/pages/temporary-accommodation/manage/bedspaceEditable.ts
```diff
@@ -12,4 +12,8 @@ export default abstract class BedspaceEditablePage extends Page {
 
     this.clickSubmit()
   }
+
+  getCharacteristics(alias: string): void {
```
patrickjfl: If this method is only ever likely to assign something to the alias 'characteristics', do we need the alias parameter? I wonder if it's more confusing than helpful. An alternative might be to call the method assignCharacteristicsToAlias and then pass 'characteristics' in to the method called inside.

patrickjfl (approved):
One suggestion but LGTM 👍

cypress_shared/pages/temporary-accommodation/manage/bedspaceEditable.ts
bryangaledxw: I think I'd rather pass in the alias - keep the alias name knowledge in one place (in the step definitions, rather than both the step definitions and the page helper class). But I think changing "get" to "assign" is a good call. How would you feel about just `assignCharacteristics(alias: string)` and the "ToAlias" is implicit by context/convention?

patrickjfl: yep sure, that makes sense :)

PR #297, opened 2023-04-18
Reallocate errors for forward compatibility with upcoming API changes

bryangaledxw:
# Changes in this PR

See #259. As part of upcoming changes to how we handle PDUs, the API will look for a `probationDeliveryUnitId` field on our `NewPremises` and `UpdatePremises` objects. If this it not found, it will use the existing `pdu` field. If *both* are not found, it wil return an "empty" error against the `probationDeliveryUnitId` field field.

To ensure the UI is compatible with these changes we reallocate errors on `probationDeliveryUnitId` field to `pdu`. This can be removed once #259 is merged in

# Release checklist

[Release process

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

- [ ] [Manually
  release to test
- [ ] [Manually
  release to preprod
- [ ] [Manually
  release to prod

### DIFF
--- a/server/@types/express/index.d.ts
+++ b/server/@types/express/index.d.ts
@@ -1,5 +1,5 @@
-import type { ErrorMessages } from '@approved-premises/ui'
 import type { ApprovedPremisesApplication as Application, ProbationRegion } from '@approved-premises/api'
+import type { ErrorMessages } from '@approved-premises/ui'
 
 export default {}
 
--- a/server/@types/express/index.d.ts
+++ b/server/@types/express/index.d.ts
@@ -33,6 +33,8 @@ declare module 'express' {
   type ShowRequestHandler = TypedRequestHandler<ShowParams>
 }
 
+type FlashMessage = string | ErrorMessages | Array<ErrorSummary> | Record<string, unknown>
+
 export declare global {
   namespace Express {
     interface User {
@@ -45,7 +47,8 @@ export declare global {
       verified?: boolean
       id: string
       logout(done: (err: unknown) => void): void
-      flash(type: string, message: string | ErrorMessages | Array<ErrorSummary> | Record<string, unknown>): number
+      flash(type: string, message: FlashMessage): number
+      flash(type: string): FlashMessage
     }
   }
 }
--- a/server/utils/validation.test.ts
+++ b/server/utils/validation.test.ts
@@ -11,6 +11,7 @@ import {
   catchValidationErrorOrPropogate,
   fetchErrorsAndUserInput,
   insertGenericError,
+  reallocateErrors,
   setUserInput,
 } from './validation'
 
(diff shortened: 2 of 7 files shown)


bryangaledxw:
@tahb Would be good to get your take on this, just want to check I'm not massively overthinking this. @danlivings-dxw Would be good to get your thumbs up on this too, that at a high level it makes sense. Last two commits are probably the ones you want to care about

bryangaledxw:

bryangaledxw:
[Update bedspace search to only present PDUs associated with user's region](https://trello.com/c/AXdBg54M/972-update-bedspace-search-to-only-present-pdus-associated-with-users-region)

server/@types/express/index.d.ts
```diff
@@ -45,7 +47,8 @@ export declare global {
       verified?: boolean
       id: string
       logout(done: (err: unknown) => void): void
-      flash(type: string, message: string | ErrorMessages | Array<ErrorSummary> | Record<string, unknown>): number
+      flash(type: string, message: FlashMessage): number
+      flash(type: string): FlashMessage
```
patrickjfl, line 51: what does it mean to have two methods with the same name on an interface? Are both signatures accepted in all cases?

server/utils/validation.ts
```diff
@@ -103,6 +103,26 @@ const extractValidationErrors = (error: SanitisedError | Error, context: ErrorCo
   throw error
 }
 
+export const reallocateErrors = (error: SanitisedError | Error, source: string, destination: string) => {
+  if (isAnnotedError(error)) {
```
patrickjfl: typo here

server/utils/validation.test.ts
```diff
@@ -312,3 +313,49 @@ describe('insertGenericError', () => {
...
+      data: {
+        'invalid-params': [
+          { propertyName: '$.uiProperty', errorType: 'someErrorType' },
+          { propertyName: '$.unrelatedProperty', errorType: 'somOtherErrorType' },
+        ],
+      },
+    })
+  })
+
+  it('does nothing if the desination parameter already exists', () => {
```
patrickjfl, line 340: I'm a bit confused here - why do we not want the apiProperty error to be reallocated?

patrickjfl (commented):
I've got no major problem but have a couple of questions to help my understanding.

server/utils/validation.test.ts
bryangaledxw: It's a bit academic as we're not expecting to get errors on both `pdu` and `probationDeliveryUnitId`, so for now this case shouldn't come up. But in general I want to err on the side of caution and not overwrite an existing error

server/utils/validation.ts
bryangaledxw: Ah, cheers!

server/utils/validation.test.ts
patrickjfl: ok yeah, that makes sense 👍

server/@types/express/index.d.ts
bryangaledxw: Yep, so `flash` can be called with either of those signature. Shows up in VS code as this
image

It expands out to 6 possibilities due to `FlashMessage` being a composite type

patrickjfl: nice

patrickjfl (approved):

PR #300, opened 2023-04-18
Upgrade to Redis 7.0.11 in dev

tahb:
- The app appears to continue working with the new redis without requiring package updates, we can sign in and that requires redis as it's the session store.
- Our environments are currently running version 4.x.x
- We've been asked to upgrade due to EOL
- I can't find out why we are using bitnami rather than the official redis image, to avoid potential problems with the upgrade I'm going to keep the surface area of problems smaller by staying with the same maintainer - one for later
- apparently redis supports backwards compatibility even between major changes[1] so we should be able to attempt the otherwise big jump in one go

[1] https://user-guide.cloud-platform.service.justice.gov.uk/documentation/deploying-an-app/redis/upgrade.html#upgrade-paths

### DIFF
--- a/docker-compose-test.yml
+++ b/docker-compose-test.yml
@@ -5,6 +5,8 @@ services:
     image: 'bitnami/redis:7.0.11'
     networks:
       - hmpps_int
+    environment:
+      - ALLOW_EMPTY_PASSWORD=yes
     ports:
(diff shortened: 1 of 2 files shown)


pezholio (approved):
LGTM 👍

PR #304, opened 2023-04-19
Update bedspace search E2E test

bryangaledxw:
We update the bedspace search E2E test, as a failure to properly read the PDU of an exiting premises is leading to a later test failure


patrickjfl (approved):
Nice one 👍

PR #307, opened 2023-04-19
Run E2E tests sequentially to avoid flakey auth failures

tahb:
# Context

# Changes in this PR

After investigating recent flakey e2e tests[2] I could see that 2 tests were failing for auth related reasons. Firstly the user is thrown to the sign out page and secondly the user gets a not authorised error. We don't always see these and when we re-run we are usually ok. We have been toying with the idea of adding retry: 1 onto all tests.

In this case we found that there was another e2e test running[3] within the same window. Given that these tests run against the same environment and sign in with the same user, I suspect that there may be odd situations where one test signs out and the other expects them to still be signed in. This would explain the lack of pattern to the problem.

In order to solve this we use a CircleCI trick by asking it to use a small resource. Only 1 small resource can be used at a time which creates a situation where these jobs cannot be parallelised.

- When the API runs e2e tests for the frontend, it already uses this to ensure the jobs are queued[1]
- CircleCI is on the performance plan that allows this resource configuration

[1] https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/.circleci/config.yml#L18
[2] https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-temporary-accommodation-ui/1689/workflows/732f7a22-c9e4-4692-84af-4e4ba17613b5/jobs/3667/artifacts
[3] https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-temporary-accommodation-ui/1684/workflows/b6e3175d-9572-4fda-aef9-761c93dc9f36

# Release checklist

[Release process

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

- [ ] [Manually
  release to test
- [ ] [Manually
  release to preprod
- [ ] [Manually
  release to prod

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,6 +44,7 @@ jobs:
           template: basic_fail_1
 
   e2e_environment_test:
+    resource_class: small
     circleci_ip_ranges: true # opt-in to jobs running on a restricted set of IPs
     docker:
       - image: cypress/browsers:node16.14.2-slim-chrome100-ff99-edge


bryangaledxw (approved):

bryangaledxw:
Seems to be causing some issues on CircleCI? 
image

tahb:
Yeah I put it in the wrong place in an earlier version, hopefully fixed now. Running CI once more after rebasing so it should be green once again 🤞

PR #312, opened 2023-04-20
Configure Cypress to not keep E2E tests in memory

bryangaledxw:
In our E2E test config, we set numTestsKeptInMemory to zero, to prevent our test environment running out of memory


richpjames (approved):

bryangaledxw:
@richpjames This was done as an emergency fix, but is probably a good thing in any case if you want to copy this? There's no value that I can see to Cypress keeping past tests in memory on CI runs

PR #342, opened 2023-05-04
Fix argument ordering in PremisesShowPage.parsePremises()

bryangaledxw:
Fix argument ordering, that was causing our E2E tests to attempt to parse the PDU as a status


patrickjfl (approved):
LGTM

PR #371, opened 2023-05-18
Copy tail-logs script from CAS1

bryangaledxw:
Update tail-logs script to use correct namespace and pod names

### DIFF
--- a/script/tail-logs
+++ b/script/tail-logs
@@ -25,9 +25,9 @@ else
 fi
 
 if [ "$app" = "api" ]; then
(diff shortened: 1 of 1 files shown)


tahb (approved):
Oh, I didn't know we had this script. I've been following [the docs](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4154196024/Tailing+Logs) which does the same thing quite easily. Not convinced we need the script but whilst we have it this correction makes sense.

PR #372, opened 2023-05-19
Present BedspaceSearchResults as a table

tahb:
# Context

https://trello.com/c/SR3qJg12

Users were reporting that they were clicking "View" links and were being taken to an unexpected bedspace. On review, this was due to the fact users were clicking the "View" link below the address, rather than the one above that was inline with the bedspace name. 

This change adds a presenter ontop of the BedspaceResults to convert them into what we need to create a govuk table. The presenter pattern wasn't there so I realise I'm suggesting this but there may be a more conventional way to do this in the project at this point? 

I noticed that [the BookingSearchService had decided to put the presentational logic within the service itself](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/main/server/services/bookingSearchService.ts#L17). For me good services have single responsibility and can be easily reused. For it to know so much about how we want to present our data felt off? If anything I'd like to change that controller action to `search` and use another presenter in the same way as this.

# Changes in this PR

### Before

Screenshot 2023-05-19 at 19-01-29 Temporary Accommodation - Find a bedspace

### After

Screenshot 2023-05-19 at 19-01-11 Temporary Accommodation - Find a bedspace

When there are no results we now hide the table headers too:

Screenshot 2023-05-22 at 16-46-35 Temporary Accommodation - Find a bedspace

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod


server/presenters/bedspaceResultTablePresenter.test.ts
```diff
@@ -0,0 +1,33 @@
+import { bedSearchResultsFactory } from '../testutils/factories'
+import { BedspaceResultTablePresenter } from './bedspaceResultTablePresenter'
+
+describe('BedspaceResultTablePresenter', () => {
+  it('present', () => {
```
bryangaledxw: Convention has been:

```
describe('ClassName', () => {
  describe('fuctionName', () => {
    it('does the thing', () => {
```

server/presenters/bedspaceResultTablePresenter.ts
```diff
@@ -0,0 +1,26 @@
+import { BedSearchResult } from '@approved-premises/api'
+import paths from '../paths/temporary-accommodation/manage'
+
+export class BedspaceResultTablePresenter {
+  static present(results: BedSearchResult[]): ({ text: string } | { html: string })[][] {
```
bryangaledxw: You should be able to use the TableRow type here

server/presenters/bedspaceResultTablePresenter.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    const searchResults = bedSearchResultsFactory.build()
```
bryangaledxw: You could do `const results = bedSearchResultFactory.buildList(2)`, and not even bother using `bedSearchResult*s*Factory`, then you can drop the `arrayContaining`

cypress_shared/pages/temporary-accommodation/manage/bedspaceSearch.ts
```diff
@@ -53,6 +53,6 @@ export default class BedspaceSearchPage extends Page {
   }
 
   clickBedspaceLink(room: Room) {
-    cy.get('h4').contains(room.name).parent().siblings().contains('View').click()
+    cy.get('td').contains(room.name).parent().find('td:contains("View")').click()
```
bryangaledxw: I've not seen this pattern before - any reason not to use something like `....find('td').contains('View').click()` as elsewhere?

server/controllers/temporary-accommodation/manage/bedspaceSearchController.ts
```diff
@@ -40,11 +41,14 @@ export default class BedspaceSearchController {
             startDate,
             durationDays,
           } as BedSearchParameters)
+
+          tableRows = BedspaceResultTablePresenter.present(results.results)
```
bryangaledxw: I feel like I'd like to discuss before taking this approach - feels like it's introducing a third way of doing things

- Doing formatting within a service, called from the controller
- Exposing functions like thing to Nunjucks and calling them directly within the template (in CAS3 we've done this for self-contained components, CAS1 has moved towards this a bit more extensively - see `nunjucksSetup.ts` in each)
- Doing formatting within a presenter, called from the controller

My feeling is I'd rather we do option 2 here (and it feels tidier to me to only have the one thing we're passing in that may-or-may-not be null depending on whether a search has been entered), but lmk what you think

With option 2, we'd have something like:

```
{{ govukTable({
  head: tableHeadings,
  rows:  BedspaceSearchResultUtils.resultTableRows(results)
}) }}
```
with the controller staying unchanged

cypress_shared/components/bedspaceSearchResult.ts
```diff
@@ -9,12 +9,12 @@ export default class BedspaceSearchResult extends Component {
   shouldShowResult(checkCount = true): void {
     cy.get('div[data-cy-result]')
```
bryangaledxw: Do we want/need this any more? Or should we just find the table?

bryangaledxw (commented):
Looks good, main thing for me is using presenter in the controller, vs. calling a similar function from the nunjucks template. Let me know what you think

server/presenters/bedspaceResultTablePresenter.test.ts
tahb: That's definitely what I wanted to do 👍

cypress_shared/pages/temporary-accommodation/manage/bedspaceSearch.ts
tahb: Ah yes, I think I was trying alternatives because the test suite started reporting that there were no tests around this time. I wrongly assumed this latest change was to blame and forgot to put it back.

cypress_shared/components/bedspaceSearchResult.ts
tahb: I thought about this. We could just match on `table` though I figured this was a convention to prevent fragile matchers should there be multiple tables on the same page? I'll remove it and we can cross with that bridge later if we come to it.

server/controllers/temporary-accommodation/manage/bedspaceSearchController.ts
tahb: I was expecting this and agree that it's totally resonable not to introduce another way of doing things. I'll take a look at moving to option 2.

server/views/temporary-accommodation/bedspace-search/index.njk
```diff
@@ -91,7 +91,7 @@
 
         {{ govukTable({
           head: tableHeadings,
-          rows: tableRows
+          rows: BedspaceSearchResultUtils.resultTableRows(results.results)
```
tahb, line 94: @bryangaledxw before I redo my commits, is this approach right? We use a module from utils instead of a presenter. We also clean up the `bedspaceSearchResult` component that we no longer need which I missed on my first pass.

bryangaledxw: Yep, looks good to me. This is definitely making me feel conscious of the naming - going with the very generic "Utils" - but it feels like it's bedded in as convention at this point

bryangaledxw (approved):
LGTM

PR #385, opened 2023-05-25
Revert to using esbuild for E2E test packaging

bryangaledxw:
We previously switched from esbuild to browserify for bundling our E2E tests for running in the test runner (see PR #164), due to an upcoming incompatibility between esbuild and cucumber (see https://github.com/badeball/cypress-cucumber-preprocessor/issues/929).

We now find that since moving to TypeScript 5, browserify is struggling with tests that reference code that uses decorators. As the esbuild/cucumber incompatibility has been resolved, we move back to using esbuild to avoid this issue.


bryangaledxw:
[Revert to using esbuild for E2E test packaging](https://trello.com/c/XmTBl0EV/1163-revert-to-using-esbuild-for-e2e-test-packaging)

tahb (approved):

PR #405, opened 2023-06-07
Only ask HMPPS Auth for the current user on a new session

tahb:
# Context

Previously we were calling the `GET /auth/api/user/me` endpoint in HMPPS Auth as part of every request through [populateCurrentUser](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/f6a0e33116a137d1de4aeebd88cdee8d92cd96c0/server/middleware/populateCurrentUser.ts#L11). I think the existing need for this is to get the user 'displayName' for the page headers and potentially sending data to HMPPS Audit.

We have seen in production that we have requests that are taking 10-30 seconds and timing out due to closed connections to HMPPS Auth. It seems like there are 3 retries each allowed to take 10 seconds. In [the top 3 slowest requests reported by the frontend](https://ministryofjustice.sentry.io/performance/?end=2023-06-07T19%3A19%3A43&project=4504129156218880&query=&sort=-p95&start=2023-05-31T19%3A19%3A43) we can see that requests are slow, retried or resulted in 3 failed retries and failed (on one occurence).

Screenshot 2023-06-07 at 12 15 15
 
Before we escalate this availability issue to HMPPS Auth, I think we have an opportunity to hit their endpoint once per session rather than once per request. This should improve our performance and reduce load on their service.

# Changes in this PR

## Screenshots of UI changes

I've attached some local frontend logs showing the logs generated when we call `userService.getUser` which demonstrates how we're now making less calls.

### Before

Screenshot 2023-06-07 at 19 08 55

### After

Screenshot 2023-06-07 at 19 10 53

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/server/middleware/populateCurrentUser.ts
+++ b/server/middleware/populateCurrentUser.ts
@@ -8,10 +8,11 @@ export default function populateCurrentUser(userService: UserService): RequestHa
     try {
       if (req.user) {
         const callConfig = extractCallConfig(req)
-        const user = res.locals.user && (await userService.getUser(callConfig))
-        if (user) {
-          res.locals.user = { ...user, ...res.locals.user }
-        } else {
+        const userDetails = req.session.userDetails || (await userService.getUser(callConfig))
+        req.session.userDetails = userDetails
+        res.locals.user = { ...userDetails, ...res.locals.user }
+
+        if (!userDetails) {
           logger.info('No user available')
         }
       }
--- a/server/@types/express/index.d.ts
+++ b/server/@types/express/index.d.ts
@@ -11,6 +11,7 @@ declare module 'express-session' {
     application: Application
     previousPage: string
     probationRegion: ProbationRegion
+    userDetails: UserDetails
   }
 }
 


server/middleware/populateCurrentUser.ts
```diff
@@ -8,9 +8,12 @@ export default function populateCurrentUser(userService: UserService): RequestHa
...
         const callConfig = extractCallConfig(req)
-        const user = res.locals.user && (await userService.getUser(callConfig))
-        if (user) {
-          res.locals.user = { ...user, ...res.locals.user }
+        const userDetails = req.session.userDetails
+          ? req.session.userDetails
+          : res.locals.user && (await userService.getUser(callConfig))
+        if (userDetails) {
+          req.session.userDetails = userDetails
+          res.locals.user = { ...userDetails, ...res.locals.user }
```
bryangaledxw: Up to you if you want to brave it, but I reckon this can be simplified a lot? I think a lot of this checking is redundant. Can it be?:

```
if (req.user) {
   const userDetails = req.session.userDetails || await userService.getUser(callConfig)
   req.session.userDetails = userDetails
   res.locals.user = { ...userDetails, ...res.locals.user }
}
```

bryangaledxw (approved):
LGTM

This is reminding of something on the BEIS project - that similarly only fetched user info at the start of the session, and when users were trying out permissions, that the target user needed to log in and out again for those new permissions to apply was a source of confusion. So I think be prepared for that if userService.getUser() is where the frontend is getting its permissions

pezholio:
Oh, this is interesting. Might be something worth nicking for CAS1 too 👍

tahb:
@bryangaledxw Good point on permissions. To get my thoughts out of my head:

I noticed we have [an unused method called getUserRoles](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/f6a0e33116a137d1de4aeebd88cdee8d92cd96c0/server/data/hmppsAuthClient.ts#L55) which gets roles from a different endpoint. 

Adam and I are thinking that in the short term we will be getting roles+permissions from the API. Then _if_ we get time we'll change where we get roles (not permissions) from so they come from Delius (probably with this unused method) instead. In both cases finer grain permissions should always be coming from the API and we can choose if we want to put them in the session or not if users have a bad experience.

I suspect we might end up with the following in our middleware:

1. On each request check the token from the Token API (nothing changes from what we have now)
2. On each session get the user details from Delius (this PR)
3. On each session or request get the use roles from the API (a new request to the API)
4. On each session or request get the user permissions from the API (a new request to the API)

Then number 3 gets swapped out to:

3. On each session get the user roles from Delius (start using getUserRoles when they exist in Delius)

server/middleware/populateCurrentUser.ts
tahb: That is much prettier, I'll give it a try thanks :)

bryangaledxw: Cool. Worth getting a test in for this? It was grandfathered in from the HMPPS Typescript Template without tests, but if we're making significant changes, might be worth testing

tahb: Your version works great so I've switched to it. The old code was logging when there were no details so I've opted to keep it just in case it's ever useful for debugging.

PR #412, opened 2023-06-12
Update getTaskStatus

bryangaledxw:
Previously getTaskStatus could return 'complete' for a task with incomplete pages, provided those pages were in a sequential run at the start of the task. We update getTaskStatus so it will return 'not_started' or 'in_progress' in this case

### DIFF
--- a/server/form-pages/utils/getTaskStatus.ts
+++ b/server/form-pages/utils/getTaskStatus.ts
@@ -1,8 +1,8 @@
-import type { Task, TaskStatus } from '@approved-premises/ui'
 import {
   TemporaryAccommodationApplication as Application,
   ApprovedPremisesAssessment as Assessment,
 } from '@approved-premises/api'
+import type { Task, TaskStatus } from '@approved-premises/ui'
 import { TasklistPageInterface } from '../tasklistPage'
 
 const getPageData = (applicationOrAssessment: Application | Assessment, taskName: string, pageName: string) => {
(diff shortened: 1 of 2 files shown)


bryangaledxw:
[Fix getTaskStatus](https://trello.com/c/sy6dLl57/1226-fix-gettaskstatus)

server/form-pages/utils/getTaskStatus.ts
```diff
@@ -1,61 +1,40 @@
...
-    const errors = page.errors()
-    // And the next page ID
-    pageId = page.next()
-
-    if (errors.length) {
-      // Are there any errors? Then the task is incomplete
-      status = 'in_progress'
-      break
+    // If there's errors for this page, then we know the task incomplete
+    if (page.errors().length) {
```
tahb, line 30: Surprised to see this but I see it existing before and there's probably a gap in my understanding. Why do pages keep track of their errors? I'm imagining that we use this method on the task list page. When users come and go from their tasks that we either accepted their previous answers to and saved them or we rejected them in which case they start again from a blank slate.

tahb (approved):
Good to have this 👍

server/form-pages/utils/getTaskStatus.ts
bryangaledxw: When filling out the form, Page.errors() is mostly used for creating these errors, that are then passed into the njk template for that page. Or if the returned array is empty, we progress the user to the next page, so we're already using `page.errors().length === 0` as proxy for "this page is complete".

localhost_3000_applications_16c791d3-ea91-47ad-84ed-fbee220d4d89_tasks_contact-details_pages_probation-practitioner

Here, read it as, "given the data we've just passed into the constructor, is this page complete?". Though, I'm struggling to get this check ever be true in practice, since we already have the `if (!pageData)` check before this. It would be hit though if we ever add a new compulsory field to a page. In that case, there will be data for this page, but we would expect errors.

PR #416, opened 2023-06-13
Check Your Answers handles branching tasks

bryangaledxw:
# Changes in this PR

This PR fixes a bug that could occur if the user completes the application form, but then changes their answers in such a way that a previously completed page is skipped. That skipped page would appear in both the check your answers screen, and in the generated document for the submitted application.

This PR addresses this by adding a `forPagesInTask()` function, which iterates through the pages of a given task, in the order a user would encounter them, and skipping any the user would skip, given their answers. For each page, a callback is called with the page object. Use of `forPagesInTask()` replaces iterating through the data on the application and trusting it not to contain any unwanted data

### Before
localhost_3000_applications_0d0244e3-9326-4260-a5e9-ed77f3e458c6_tasks_check-your-answers_pages_review (1)

Note that answers regarding DTR are shown, despite the user answering "no" to the first question, and skipping this page

### After
localhost_3000_applications_0d0244e3-9326-4260-a5e9-ed77f3e458c6_tasks_check-your-answers_pages_review

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/server/services/tasklistService.test.ts
+++ b/server/services/tasklistService.test.ts
@@ -1,51 +1,63 @@
 import getTaskStatus from '../form-pages/utils/getTaskStatus'
 import { applicationFactory } from '../testutils/factories'
+import getSections from '../utils/assessments/getSections'
 import TasklistService from './tasklistService'
 
-jest.mock('../form-pages/apply', () => {
-  return {
-    sections: [
+jest.mock('../form-pages/utils/getTaskStatus')
+jest.mock('../utils/assessments/getSections')
+
+const sections = [
+  {
+    title: 'First section',
+    name: 'first-section',
+    actionText: 'Complete  section',
+    tasks: [
       {
-        title: 'First Section',
-        tasks: [
-          {
-            id: 'first-task',
-            title: 'First task',
-          },
-          {
-            id: 'second-task',
-            title: 'Second task',
-          },
-        ],
+        id: 'first-task',
+        title: 'First task',
+        actionText: 'Complete first task',
+        pages: {},
       },
       {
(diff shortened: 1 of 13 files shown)


bryangaledxw:
[Check Your Answers handles branching tasks](https://trello.com/c/CwMXhaNT/1191-check-your-answers-handles-branching-tasks)

tahb (approved):
I think we should get this in as we need a fix to this problem 👍 

Slightly worried about the number of loops we have to go through in terms of minor performance, increased complexity and a need to remember to use the right methods when submitting and on CYA to force these checks.  
 
Were there any other approaches you considered and rejected? I'm wondering if it would be an idea to hook into the update of a page that removes the need for an old branch. Rather than re-establishing the integrity of all the pages when handling the application we could just re-establish the integrity of all pages within that same task.
  
I also wondered when reading this if it'd be easier to drop support for assessments, though the refactoring here seems really positive. Maybe one for the future.

bryangaledxw:
"I also wondered when reading this if it'd be easier to drop support for assessments" - yeah, I've tried to tuck assessments more out-of-the-way with the change to getSections, but I didn't want to get drawn into a total removal

"Were there any other approaches you considered and rejected? " - I did wonder about doing this on a per-task basis, but that would have been a far larger upheaval to the existing system. And I feel like it's an improvement that both getResponses() and getTaskResponsesAsSummaryListItems() are using `Apply` as the source of truth of what tasks to expect, rather than the mutable application data

PR #420, opened 2023-06-14
Improve application stubbing

bryangaledxw:
We add stubs for completed applications, so that when running the UI in isolation from the API, we can easily reach a desired Apply page rather than needing to complete the entire application first


bryangaledxw:
[Improve application stubbing](https://trello.com/c/7qVFwUB1/1241-improved-application-stubbing)

tahb (approved):

PR #438, opened 2023-06-23
Add Prison information screens

bryangaledxw:
# Changes in this PR

This PR adds two new pages for fetching information from NOMIS and including it in the Apply form.

## Screenshots of UI changes
localhost_3000_applications_4f804e4c-3887-4fd7-a8cc-58d68399ffd3
localhost_3000_applications_4f804e4c-3887-4fd7-a8cc-58d68399ffd3_tasks_prison-information_pages_adjudications
localhost_3000_applications_4f804e4c-3887-4fd7-a8cc-58d68399ffd3_tasks_prison-information_pages_acct-alerts
localhost_3000_applications_4f804e4c-3887-4fd7-a8cc-58d68399ffd3_tasks_check-your-answers_pages_review
localhost_3000_applications_4f804e4c-3887-4fd7-a8cc-58d68399ffd3_tasks_prison-information_pages_adjudications (1)
localhost_3000_applications_4f804e4c-3887-4fd7-a8cc-58d68399ffd3_tasks_prison-information_pages_acct-alerts (1)
localhost_3000_applications_4f804e4c-3887-4fd7-a8cc-58d68399ffd3_tasks_check-your-answers_pages_review (1)

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod


bryangaledxw:
[Add Prison information screens](https://trello.com/c/8qYaSMjX/1253-add-prison-information-screens)

selina-dxw:
On the ACCT screen underneath the navigation bar can we include the full name for ACCT. For example: 

"Assessment, Care in Custody and Teamwork (ACCT)"

And can we change the button text from 'Save and continue' to 'Continue'

Can we also include guidance on this screen above the navigation bar:

"Review the adjudication and ACCT information provided below to help inform your referral. 

This information will be included as part of your referral."

This guidance might change depending on how we decide to handle this screen but that's probably good enough for now.

bryangaledxw:
@selina-dxw Do we definitely want this to be "continue"? When I flagged with Barry that this does save data imported data into our form, he said in that case to go with "Save and continue" - https://mojdt.slack.com/archives/C052G56JS5P/p1687517171640279

selina-dxw:
Ah sorry I forgot about that, yep stick with 'Save and continue'. I thought about changing the guidance but I think it's better to keep it simple for now.

bryangaledxw:
@selina-dxw Done ✅

server/form-pages/apply/risk-strength-and-need-information/prison-information/adjudications.ts
```diff
@@ -0,0 +1,76 @@
...
+import type { DataServices, PageResponse } from '@approved-premises/ui'
+import { CallConfig } from '../../../../data/restClient'
+import { DateFormats } from '../../../../utils/dateUtils'
+import { sentenceCase } from '../../../../utils/utils'
+import TasklistPage from '../../../tasklistPage'
+import { Page } from '../../../utils/decorators'
+
+type AdjudicationsBody = {
+  adjudications: Array<
+    Omit<Adjudication, 'finding'> & {
```
tahb: Presumably this is modifying the original Adjudication type and we can't or don't want to make this change on the main Adjudication. I'm thinking that if it can be null, should we not make this change globally.

server/form-pages/apply/risk-strength-and-need-information/prison-information/adjudications.ts
```diff
@@ -0,0 +1,76 @@
...
+  next() {
+    return 'acct-alerts'
+  }
+
+  response() {
+    const response: PageResponse = {}
+
+    response.Adjudications = this.body.adjudications.length
+      ? this.body.adjudications.map(adjudicationResponse)
+      : `No adjudication information available for ${this.application.person.name}`
```
tahb: If there was no info found and users could look back at old referrals they might wonder if this is being pulled in live. Perhaps `No adjudication information was available for ${this.application.person.name} at the time of referral` would be clearer.

tahb (approved):
Nice 👍 A couple of non blocking comments

server/form-pages/apply/risk-strength-and-need-information/prison-information/adjudications.ts
bryangaledxw: The significant bit here isn't `| null`, it's `/** @nullable */`. That's what gets us the schema we want, but even explicitly marking a field as nullable in the `api.yml` file doesn't get us the `/** @nullable */` comment in the generated types.

tbh, the more I think about this the less I like it - if the API changes and we get extra data back from the server, that's automatically going to be included, and will potentially cause the schema validation to fail. Either we use the "all" schema and this is moot, or we should do some preprocessing on the adjudications before including in our form data

bryangaledxw: I've taken latter approach - rather than bodging the API types with extra declarations, there's a method mapping the data we get from the API to fixed UI types with stronger guarantees about what will be present

PR #450, opened 2023-06-29
Arrivals API Change

bryangaledxw:
We now need to send a `type` field when creating an arrival. We update to the latest shared types, and update `ArrivalsController` to set this new field


richpjames (approved):

server/controllers/temporary-accommodation/manage/arrivalsController.ts
```diff
@@ -55,6 +55,7 @@ export default class ArrivalsController {
         ...req.body,
         ...DateFormats.dateAndTimeInputsToIsoString(req.body, 'arrivalDate'),
         ...DateFormats.dateAndTimeInputsToIsoString(req.body, 'expectedDepartureDate'),
+        type: 'CAS3',
```
pezholio, line 58: Wondering if this would be better placed in the client, rather than the controller? No biggie if you think otherwise tho 👍

pezholio (approved):

server/controllers/temporary-accommodation/manage/arrivalsController.ts
bryangaledxw: I'd thought maybe service, wanting to keep the client fairly naive, and this feeling lower level than the controller. but then I'm fiddling about with `Omit<NewArrival, 'type'>` in the controller to keep the types honest 😕

PR #501, opened 2023-07-26
(1349) Assessor can make state changes to application

richpjames:
# Context
An assessment needs to be able to move between the states outlined in this diagram (with the adjustment that from the rejected state we can only transition to unallocated).
image

In order to do this we need to:
- be able to 'show' an assessment
- be able to update the state to one of the viable states, given the state we're in 
- confirm this transition
- perform it
- show the user confirmation that the assessment has been updated

This work builds on the the `feature/1346-an-assessor-can-see-an-application-listing-page-from-the-dashboard` branch which should be merged first

[Trello card](https://trello.com/c/ZI5W46gr/1349-an-assessor-can-make-state-changes-to-an-application)

# Changes in this PR
- We update the Assessments used throughout the application to the correct domain: `TemporaryAccommodationAssessment`
- We add a temporary type for the `TemporaryAccommodationAssessment` with the desired statuses as this has not yet been added in the [api spec](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/src/main/resources/static/api.yml). This will need to be deleted once the spec is correct
- We add the 'find' method to the assessment service
- We add some utils for generating the 'status' badge and 'actions' menu on the assessments/show view
- We add the 'show' view 
- We add the confirm view
- We add the client, service and controller methods to update the assessment
- We extend the intergration tests

## Screenshots of UI changes

Apply -- Assess -- I can change the state of the referral from unallocated to in review (2)
confirm-in-review (11)
confirm-ready-to-place
confirm-close
show-with-banner

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/integration_tests/tests/temporary-accommodation/assess/assess.cy.ts
+++ b/integration_tests/tests/temporary-accommodation/assess/assess.cy.ts
@@ -16,47 +16,49 @@ context('Apply', () => {
   })
 
   describe('Assess', () => {
-    it('I can view a list of assessments', () => {
-      // Given there are assessments in the database
-      const inProgressAssessmentSummaries = assessmentSummaryFactory.buildList(2, { status: 'in_review' })
-      const unallocatedAssessmentSummaries = assessmentSummaryFactory.buildList(2, { status: 'unallocated' })
-      const readyToPlaceAssessmentSummaries = assessmentSummaryFactory.buildList(2, { status: 'ready_to_place' })
-      const rejectedAssessmentSummaries = assessmentSummaryFactory.buildList(2, { status: 'rejected' })
-      const closedAssessmentSummaries = assessmentSummaryFactory.buildList(2, { status: 'closed' })
+    describe('List assessments', () => {
+      it('I can view a list of assessments', () => {
+        // Given there are assessments in the database
+        const inProgressAssessmentSummaries = assessmentSummaryFactory.buildList(2, { status: 'in_review' })
+        const unallocatedAssessmentSummaries = assessmentSummaryFactory.buildList(2, { status: 'unallocated' })
+        const readyToPlaceAssessmentSummaries = assessmentSummaryFactory.buildList(2, { status: 'ready_to_place' })
+        const rejectedAssessmentSummaries = assessmentSummaryFactory.buildList(2, { status: 'rejected' })
+        const closedAssessmentSummaries = assessmentSummaryFactory.buildList(2, { status: 'closed' })
 
-      cy.task('stubAssessments', [
-        ...inProgressAssessmentSummaries,
-        ...unallocatedAssessmentSummaries,
-        ...readyToPlaceAssessmentSummaries,
-        ...rejectedAssessmentSummaries,
-        ...closedAssessmentSummaries,
-      ])
+        cy.task('stubAssessments', [
+          ...inProgressAssessmentSummaries,
+          ...unallocatedAssessmentSummaries,
+          ...readyToPlaceAssessmentSummaries,
+          ...rejectedAssessmentSummaries,
+          ...closedAssessmentSummaries,
+        ])
 
-      // When I visit the dashboard
-      const dashboardPage = DashboardPage.visit()
+        // When I visit the dashboard
+        const dashboardPage = DashboardPage.visit()
 
-      // And I click on the "Review and assess referrals" link
-      dashboardPage.clickReviewAndAssessReferrals()
+        // And I click on the "Review and assess referrals" link
+        dashboardPage.clickReviewAndAssessReferrals()
 
-      // Then I should see the list of referrals
-      const listPage = Page.verifyOnPage(
-        ListPage,
-        unallocatedAssessmentSummaries,
-        inProgressAssessmentSummaries,
-        readyToPlaceAssessmentSummaries,
-        [...rejectedAssessmentSummaries, ...closedAssessmentSummaries],
-      )
+        // Then I should see the list of referrals
+        const listPage = Page.verifyOnPage(
+          ListPage,
+          unallocatedAssessmentSummaries,
+          inProgressAssessmentSummaries,
+          readyToPlaceAssessmentSummaries,
+          [...rejectedAssessmentSummaries, ...closedAssessmentSummaries],
+        )
 
-      // And I should see the list of referrals
-      listPage.shouldShowUnallocatedAssessments()
-      listPage.shouldShowInProgressAssessments()
-      listPage.shouldShowReadyToPlaceAssessments()
+        // And I should see the list of referrals
+        listPage.shouldShowUnallocatedAssessments()
+        listPage.shouldShowInProgressAssessments()
+        listPage.shouldShowReadyToPlaceAssessments()
 
-      // Given I click on the 'View archived assessments' link
-      listPage.clickViewArchivedReferrals()
+        // Given I click on the 'View archived assessments' link
+        listPage.clickViewArchivedReferrals()
 
-      // Then I should see the list of archived assessments
-      listPage.shouldShowArchivedAssessments()
+        // Then I should see the list of archived assessments
+        listPage.shouldShowArchivedAssessments()
+      })
     })
   })
 })
--- a/integration_tests/tests/temporary-accommodation/assess/assess.cy.ts
+++ b/integration_tests/tests/temporary-accommodation/assess/assess.cy.ts
@@ -65,7 +65,7 @@ context('Apply', () => {
 
     describe('Show an assessment', () => {
       it('I can change the state of an assessment', () => {
-        const assessment = assessmentFactory.build({ status: 'unallocated' })
+        const assessment = assessmentFactory.build({ status: 'unallocated', decision: 'accepted' })
         const assessmentSummary = assessmentSummaryFactory.buildList(1, {
           status: 'unallocated',
           person: assessment.application.person,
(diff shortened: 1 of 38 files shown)


server/utils/assessmentsUtils.ts
```diff
@@ -74,3 +78,61 @@ const dateValue = (date: string) => {
...
+    },
+    readyToPlace: {
+      text: 'Ready to place',
+      classes: 'govuk-button--secondary',
+      href: paths.assessments.confirm({ id: assessment.id, status: 'ready_to_place' }),
+    },
+    close: {
+      text: 'Close',
+      classes: 'govuk-button--secondary',
+      href: paths.assessments.confirm({ id: assessment.id, status: 'closed' }),
```
bryangaledxw: Suggest add paths in their own commit ahead of this

server/views/temporary-accommodation/assessments/show.njk
```diff
@@ -0,0 +1,26 @@
...
+{% extends "../../partials/layout.njk" %}
+
+{% set pageTitle = applicationName + " - " + assessment.application.person.name + " referral" %}
+{% set mainClasses = "app-container govuk-body" %}
+
+{% block content %}
+  {% include "../../_messages.njk" %}
+
+  {{
+        mojIdentityBar({
```
bryangaledxw: Please indent with two spaces

server/controllers/temporary-accommodation/manage/assessmentsController.ts
```diff
@@ -40,6 +40,33 @@ export const archivedAssessmentsTableHeaders = [
...
+    title: 'Mark this referral as in review',
+    text: '<p class="govuk-body">Mark this referral as in review if you or a HPT colleague are working on this referral.</p>',
+  },
+  ready_to_place: {
+    title: 'Confirm this referral is ready to place',
+    text: `<p class="govuk-body">Mark this referral as ready to place to find a suitable bedspace.</p>
+      <p class="govuk-body">Some regions will require approval before a bedspace is booked. Other regions require an address before approvals can be given. Check with your manager if you are unsure.</p>
+      <p class="govuk-body">Once a person has been placed the referral should be archived.</p>`,
+  },
+  rejection: {
```
bryangaledxw: Should this be "rejected" rather than "rejection", to match the statuses on the type?

server/controllers/temporary-accommodation/manage/assessmentsController.ts
```diff
@@ -40,6 +40,33 @@ export const archivedAssessmentsTableHeaders = [
     },
   },
 ]
+export const confirmationPageContent: Record<string, Record<'title' | 'text', string>> = {
```
bryangaledxw: Can the type be more precise? Does:

`Record<Assessment['status'], {title: string, text: string}>`

work?

server/views/temporary-accommodation/assessments/confirm.njk
```diff
@@ -0,0 +1,22 @@
...
+  <h1 class="govuk-heading-l">{{content.title}}</h1>
+  {{content.text | safe}}
+
+  <form  action="{{ paths.assessments.update({id: id, status: status}) }}?_method=PUT" method="post">
+    <input type="hidden" name="_csrf" value="{{ csrfToken }}"/>
+
+    {{ govukButton({
+          text: "Confirm",
+          preventDoubleClick: true
+        }) }}
```
bryangaledxw: Please fix indentation

bryangaledxw (commented):
Will do another pass once the underlying PR is in, but I've made some interim comments

server/services/assessmentsService.test.ts
```diff
@@ -97,4 +97,18 @@ describe('AssessmentsService', () => {
...
   })
+
+  describe('findAssessment', () => {
+    it('calls the find method on the assessment client and returns the result', () => {
+      const assessment = assessmentFactory.build()
+      const assessmentId = 'some-id'
+
+      assessmentClient.find.mockResolvedValue(assessment)
+
+      expect(service.findAssessment(callConfig, assessmentId)).resolves.toEqual(assessment)
```
bryangaledxw: I *think* this wants an `await`

server/controllers/temporary-accommodation/manage/assessmentsController.ts
```diff
@@ -59,4 +60,18 @@ export default class AssessmentsController {
...
+  show(): RequestHandler {
+    return async (req: Request, res: Response) => {
+      const callConfig = extractCallConfig(req)
+
+      const assessment = await this.assessmentsService.findAssessment(callConfig, req.params.id)
+
+      return res.render('temporary-accommodation/assessments/show', {
+        assessment,
+        actions: assessmentActions(assessment),
+        statusBadge: statusTag(assessment.status),
```
bryangaledxw: I realise we're super inconsistent in this, which isn't helpful 😅. My feeling is that status tag is a small enough, purely presentation enough, thing that it's better to call in the njk file than the controller. My suggestion is rename the existing nunjucks `statusTag` to `personStatusTag` and expose this as 'assessmentStatusTag', rather than adding a new way of getting the statusTag into the view

integration_tests/tests/temporary-accommodation/assess/assess.cy.ts
```diff
@@ -16,47 +16,49 @@ context('Apply', () => {
   })
 
   describe('Assess', () => {
-    it('I can view a list of assessments', () => {
```
bryangaledxw, line 19: Thanks for doing this in it's own commit 👍

server/views/temporary-accommodation/assessments/show.njk
```diff
@@ -0,0 +1,26 @@
...
+    })
+  }}
+
+{% endblock %}
+
+{% block extraScripts %}
+  <script type="text/javascript" nonce="{{ cspNonce }}">
+    new MOJFrontend.ButtonMenu({container: $('.moj-button-menu'), mq: "(min-width: 200em)", buttonText: "Update referral status", menuClasses: "moj-button-menu__wrapper--right"});
+  </script>
+{% endblock %}
```
bryangaledxw, line 31: Suggest wrap this in `{% if actions %}` same as elsewhere - it's caught me off-guard before that this function causes browser-side JS errors if actions is null

server/controllers/temporary-accommodation/manage/assessmentsController.ts
```diff
@@ -30,6 +31,33 @@ export const assessmentsTableHeaders = [
   },
 ]
 
+export const confirmationPageContent: Record<Assessment['status'], Record<'title' | 'text', string>> = {
```
bryangaledxw: Few instances of ` quotes that could be plain ' quotes in this block

server/routes/temporary-accommodation/manage.ts
```diff
@@ -304,6 +304,9 @@ export default function routes(controllers: Controllers, services: Services, rou
   get(paths.assessments.show.pattern, assessmentsController.show(), {
     auditEvent: 'VIEW_ASSESSMENT',
   })
+  get(paths.assessments.confirm.pattern, assessmentsController.confirm(), {
+    auditEvent: 'CONFIRM_ASSESMENT_STATUS_CHANGE',
```
bryangaledxw: I think this wants to be something like `VIEW_ASSESSMENT_STATUS_CHANGE_CONFIRM` - generally our `get`s are VIEW_s

server/controllers/temporary-accommodation/manage/assessmentsController.ts
bryangaledxw: And I think this type can still be more precise - {title: string, text: string} says they both want to be there, and I think is more straightforward than Record<'title' | 'text', string>. And I expect there's a way to enforce that this covers all statuses (do we want this to cause a compile error if new statuses are added we're not handling here?)

server/controllers/temporary-accommodation/manage/assessmentsController.ts
```diff
@@ -112,4 +114,18 @@ export default class AssessmentsController {
...
+  update(): RequestHandler {
+    return async (req: Request, res: Response) => {
+      const callConfig = extractCallConfig(req)
+
+      const assessment = await this.assessmentsService.findAssessment(callConfig, req.params.id)
+
+      assessment.status = req.params.status as Assessment['status']
+      await this.assessmentsService.updateAssessment(callConfig, assessment)
+
+      req.flash('info', `Assessment updated status updated to "${lowerCase(assessment.status)}"`)
```
bryangaledxw: I'm a bit wary about tying user-facing naming to API naming - thought on `lowerCase(statusName(assessment.status))`?

server/routes/temporary-accommodation/manage.ts
```diff
@@ -308,5 +308,9 @@ export default function routes(controllers: Controllers, services: Services, rou
     auditEvent: 'CONFIRM_ASSESMENT_STATUS_CHANGE',
   })
 
+  put(paths.assessments.update.pattern, assessmentsController.update(), {
+    auditEvent: 'UPDATE_ASSESSMENT',
+  })
```
bryangaledxw: My feeling is use the same conventions as other `puts`, and use `redirectAuditEventSpecs`, even though we only have the one possible redirect destination

bryangaledxw (approved):
Lovely stuff! Few minor comments, but otherwise LGTM

server/controllers/temporary-accommodation/manage/assessmentsController.ts
richpjames: I've used backticks for multiline strings as I think it's more readable than concatenating the lines or having very long line, happy to change it if you disagree

server/controllers/temporary-accommodation/manage/assessmentsController.ts
richpjames: yeah makes sense

server/controllers/temporary-accommodation/manage/assessmentsController.ts
bryangaledxw: No, that's totally fair!

bryangaledxw:
@tahb Do you want to take a look at this before I merge, so I'm not approving my own changes?

Most of my changes are the first 4 commits, plus "Add state transition methods to Assessments client", and "Add 'update' method to Assessments service"

patinipat (approved):
code LGTM
still need to setup my backend to test drive though

PR #504, opened 2023-07-26
Disable API retries for PUT/POST

tahb:
# Context

# Changes in this PR

Following on with CAS1’s finding that this can create up to 3 duplicate records in the case of API failures[1].

We introduce the COMMUNITY_ACCOMMODATION_API_TIMEOUT_RESPONSE environment variable in the same way as CAS1. Right now they are all configured the same however we may want to refine this in future.

Increasing the timeout to the API to 30 seconds creates enough space for the API (who also has a 10 second timeout) to gracefully recover from a timeout to the integration service. Perhaps this should be 15-20 seconds as 30 seems a long time but I'm taking the steer from CAS1. What do we think?

[1] https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/981

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -93,10 +93,10 @@ export default {
     approvedPremises: {
       url: get('APPROVED_PREMISES_API_URL', 'http://localhost:9092', requiredInProduction),
       timeout: {
-        response: 10000,
-        deadline: 10000,
+        response: Number(get('COMMUNITY_ACCOMMODATION_API_TIMEOUT_RESPONSE', 10000)),
+        deadline: Number(get('COMMUNITY_ACCOMMODATION_API_TIMEOUT_DEADLINE', 10000)),
       },
-      agent: new AgentConfig(10000),
+      agent: new AgentConfig(Number(get('COMMUNITY_ACCOMMODATION_API_TIMEOUT_RESPONSE', 10000))),
       serviceName: 'temporary-accommodation',
     },
     audit: {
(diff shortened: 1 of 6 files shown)


server/config.ts
```diff
@@ -93,10 +93,10 @@ export default {
     approvedPremises: {
       url: get('APPROVED_PREMISES_API_URL', 'http://localhost:9092', requiredInProduction),
       timeout: {
-        response: 10000,
-        deadline: 10000,
+        response: Number(get('COMMUNITY_ACCOMMODATION_API_TIMEOUT_RESPONSE', 10000)),
```
tahb, line 96: Not sure if we need these two changed as well? CAS1 found that only changing the agent didn't have the desired effect in dev.

server/config.ts
```diff
@@ -93,10 +93,10 @@ export default {
...
+        deadline: Number(get('COMMUNITY_ACCOMMODATION_API_TIMEOUT_RESPONSE', 10000)),
```
bryangaledxw: I'd be tempted to use COMMUNITY_ACCOMMODATION_API_TIMEOUT_*DEADLINE*, to match TOKEN_VERIFICATION_API_TIMEOUT_*  and HMPPS_AUTH_TIMEOUT_*

bryangaledxw (approved):
LGTM

PR #524, opened 2023-08-02
Risk management and RoSH summary pages pull data from OASys

bryangaledxw:
# Changes in this PR

* Updates to our OASys import code, copied from CAS1
* Change to our RoSH level page - showing a warning if risk data could not be read from OASys
* Changes to our risk management plan page - getting questions and pre-filled answers from OASys, or falling back to JSON-driven questions, rather than using a static set of questions
* A versioning pattern, used on our risk management plan page

#### RoSH level page
rosh before

#### Risk management page
rmp before

#### RoSH level page
rosh after

#### Risk management page
rmp after

#### Risk management page errors
rmp after error

#### Risk management page CYA
cya after

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/server/views/applications/pages/assess-placement-risks-and-needs/placement-considerations/risk-management-plan.njk
+++ b/server/views/applications/pages/assess-placement-risks-and-needs/placement-considerations/risk-management-plan.njk
@@ -16,6 +16,8 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
 
+      <input type="hidden" name="version" value="{{ page.latestVersion }}"/>
+
       {% if page.oasysSuccess %}
         <p>Imported from OASys: <strong>{{ formatDate(page.body.oasysImported) }}</strong></p>
         <p>OASys last updated: <strong>{{ formatDate(page.body.oasysCompleted) }}</strong></p>
--- a/server/views/applications/pages/assess-placement-risks-and-needs/placement-considerations/risk-management-plan.njk
+++ b/server/views/applications/pages/assess-placement-risks-and-needs/placement-considerations/risk-management-plan.njk
@@ -1,5 +1,6 @@
 {% from "govuk/components/details/macro.njk" import govukDetails %}
 {% from "govuk/components/warning-text/macro.njk" import govukWarningText %}
+{% from "govuk/components/textarea/macro.njk" import govukTextarea %}
 {% from "../../../../components/riskWidgets/mappa-widget/macro.njk" import mappaWidget %}
 {% from "../../../../components/riskWidgets/rosh-widget/macro.njk" import roshWidget %}
 {% from "../../../../components/riskWidgets/risk-flag-widget/macro.njk" import riskFlagWidget %}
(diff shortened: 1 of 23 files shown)


bryangaledxw:
[Risk management and RoSH summary pages pull data from OASys](https://trello.com/c/XSZMaoNK/1397-risk-management-and-rosh-summary-pages-pull-data-from-oasys)

bryangaledxw:
@tahb A few things I want to flag:

* When we can't communicate with OASys, we fall back to using the questions in [oasysStubs.json](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/main/server/data/stubs/oasysStubs.json), as per CAS1. I've not delved into whether those are definitely the right questions for CAS3

* Given that getOasysSections() is necessarily called both when the page is first displayed and when the user clicks save, and given that data is effectively co-owned (imported from OASys, but editable by the user), it's quite hard to reason whether we're doing the right thing when it comes to displaying the "last updated" date. I *think* my change in my "Update dates persisted by getOasysSections()" is an improvement, but I think there are still some failure cases

* The final commit I think is a decent approach for page versioning in general, in the absence of time for a more sophisticated system

bryangaledxw:
I've also extracted this ticket - https://trello.com/c/GKSfwwdF/1430-fix-question-sorting-on-risk-management-plan-apply-screen

tahb:
I had started digging into the risk management plan questions, [discovering there would be at most 8](https://github.com/ministryofjustice/hmpps-probation-integration-services/blob/5a10c0aa4da90bd77183391921f9c927783453d7/projects/approved-premises-and-oasys/src/main/kotlin/uk/gov/justice/digital/hmpps/model/RiskManagementPlanDetails.kt#L33). I'll go and check with Selina and Barry to figure out if there's any reason not to consider all 8.

server/views/applications/pages/assess-placement-risks-and-needs/placement-considerations/risk-management-plan.njk
```diff
@@ -16,6 +16,8 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
 
+      <input type="hidden" name="version" value="{{ page.latestVersion }}"/>
```
tahb, line 19: Very nice approach. We can layer this approach into each question if/when we need to.

tahb (approved):
This looks good to me. I've got [a thread going with other folks](https://mojdt.slack.com/archives/C03GF35FY7P/p1691078467053259) to be clear about handling 8 (instead of 4) risk management questions + more if that changes in future.

I think the oasys import date choices are also reasonable and we'll see what users make of them since this has come from the central team.

tahb:
@bryangaledxw [Selina has suggested that the 4 new sections of the RMP aren't useful for our users](https://www.figma.com/file/lgKst3F6PAgmIFnEtKdfLe?node-id=2478:134759&mode=design#511327526). How much work would it be to put an allow list around the 4 questions we care about so that we forget/ignore the remaining 4 for either manual and automatic input?

bryangaledxw:
@tahb I don't think huge, if you're happy for me to spend more time on this. On Selina's point about sorting, I pulled out a ticket for that, as I didn't feel I had time to address that - https://trello.com/c/GKSfwwdF/1430-fix-question-sorting-on-risk-management-plan-apply-screen

tahb:
@bryangaledxw Agree on sorting, they should still be able to succeed in spite of that.

I've talked the idea of removing the 4 new questions through with Selina some more and I think we have to do default back to the original 4 questions please. 

The original 4 were on the old CPP referral forms and have been tested. The new questions are untested and may cause problems.

Referrals are still disabled in prod so you have the option of merging this in as is and adding a follow up PR. That may also give us an easier way to revert and add the questions all back in should we learn that HPTs really need them after all.

bryangaledxw:
@tahb Question filtering is in if you want to take another look. Relevant commit is "Filter risk management questions"

tahb (approved):
LGTM

(I do wonder if the RM question numbers will change one day. As we should fail safely it's not too big a worry)

bryangaledxw:
@tahb Yeah, I think ideal world this filtering would be API side. Checking the RM numbers seemed safer than checking the labels, and in any case, if things change, the E2Es will fail so we'll know about it

PR #535, opened 2023-08-08
A user can also link to a referral as part of creating a provisional booking

bryangaledxw:
# Changes in this PR

This PR adds a new screen as part of the booking creation journey to allow the user to associate a booking with an assessment

Where a booking is associated with an assessment, we link back to the assessment from the booking page

## To note
- This cannot be merged until API changes are made - awaiting support for linking bookings to assessments, and filtering assessments from the API by CRN
- This feature suggests a rethink of how we approach E2E tests, which is out of scope for now. See [this ticket](https://trello.com/c/bPFBAWyf/1440-receive-and-place-e2e-testing)
- This feature relies on the `applyDisabled` flag, which [this ticket](https://trello.com/c/PBk6dQWJ/1361-only-expected-regions-can-interact-with-referrals) will likely remove in favour of per-region filtering. Will need to co-ordinate these two changes
- The Figma designs show the name of the referral submitter in the referral list. We do not currently get this data from the API. This is covered by [these](https://trello.com/c/C6cxR1z9/1443-display-referrer-name-in-referral-list-when-creating-a-booking) [tickets](https://trello.com/c/Pb7gAP75/1442-include-referrer-name-when-returning-a-list-of-assessmentsummarys)

#### Select assessment
select

#### Select assessment with error
select

#### Select assessment with no assessments
select no referrals

#### Select assessments when feature is not enabled for this region
select apply disabled

#### Booking with assessment link
booking-with-link

# Release checklist

[Release process documentation](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/edit-v2/4247847062?draftShareId=a1c360ab-bd31-4db1-aae3-7cc002761de9).

## Pre-merge checklist

- [ ] Are any changes required to dependent services for this change to work?
  (eg. CAS API)
- [ ] Have they been released to production already?

## Post-merge checklist

- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to test
- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to preprod
- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to prod

PR #539, opened 2023-08-08
Fix/guard against limited person details

tahb:
# Context

# Changes in this PR

Right now if a user of CAS3 tries to view person details related to a CRN that they don't have access to (LAO) they will receive 500 errors in various places in the service. They'll be unable to create bookings or view bookings if they were created by someone else who did have access.

[The API has proposed changes to restrict person information if the logged in user shouldn't have access to a limited access offender](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/804). Instead of returning name, DOB, CRN, etc, it will now only return CRN.

These are the pages at risk of breaking badly:

1. Creating a referral when confirming the CRN
2. Viewing a full referral including CYA
3. Viewing a bedspace and all it's bookings (if one includes an LAO the page will break)
4. Viewing an individual booking
 
[CAS1 also have some more thorough changes that expect new Person types and make changes to the view by handling the different types](https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/974). This is a great long term approach and we should do this too but it's unclear to me how we'll ship both sets of changes in a backwards compatible way? **A smaller set of temporary guards that work with both the new and old version of the API is needed I think - this is the intention of this pull request.**

[For testing this locally see my comment here for how I've got the API running locally in such a way that any request for person data returns the restricted set](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/804#issuecomment-1668275157).

It seems that if we continue to expect the same `Person` type back from the API despite the API now returning `FullPerson` and `RestrictedPerson` we continue to end up with a compatible object. I may have missed something fatal here in this assumption. 

Most of the view rendering continues to work except for the person's dateOfBirth, which fails when we pass `undefined` to `formatDate`, and the person's name that doesn't cause an error but does look a bit odd when it's missing. The other field like status, gender, ethnicity, all fail gracefully and are hidden when that data isn't available to present.

From here we add a few basic guards to our views to take our pages from hard failure (current behaviour) to a degraded but functional view.

Once this is merged, and perhaps a similar PR on CAS1 I believe the [API work](https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/974) could be merged in.

### Before

On each page we through a 500, in dev we can see the following:

[new referral]
new referral
[bedspace show]
bedspace show
[booking show]
booking show
[confirm booking]
confirm booking

### After
Screenshot 2023-08-08 at 17-48-28 Temporary Accommodation - View a booking
Screenshot 2023-08-08 at 17-48-37 Temporary Accommodation - View a bedspace
Screenshot 2023-08-08 at 17-49-12 Temporary Accommodation - Confirm undefined's details

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/server/views/temporary-accommodation/components/pop-details-header/macro.njk
+++ b/server/views/temporary-accommodation/components/pop-details-header/macro.njk
@@ -4,14 +4,24 @@
 
 <div class="pop-details-header">
   {% if person %}
-
-    {% if config.nameLink %}
-      <p><span class="govuk-!-font-weight-bold">Name:</span> <a class="govuk-link" href="{{ config.nameLink }}">{{ person.name }}</a><br />
+    {% if person.name %}
+      {% if config.nameLink %}
+        <p><span class="govuk-!-font-weight-bold">Name:</span> <a class="govuk-link" href="{{ config.nameLink }}">{{ person.name }}</a><br />
+      {% else %}
+        <p><span class="govuk-!-font-weight-bold">Name:</span> {{ person.name }}<br />
+      {% endif %}
     {% else %}
-      <p><span class="govuk-!-font-weight-bold">Name:</span> {{ person.name }}<br />
+      {% if config.nameLink %}
+        <p><span class="govuk-!-font-weight-bold"><a class="govuk-link" href="{{ config.nameLink }}">Limited access offender</a></span><br />
+      {% else %}
+        <p><span class="govuk-!-font-weight-bold">Limited access offender</span><br />
+      {% endif %}
     {% endif %}
 
-    <span class="govuk-!-font-weight-bold">Date of birth:</span> {{formatDate(person.dateOfBirth)}}<br />
+    {% if person.dateOfBirth %}
+      <span class="govuk-!-font-weight-bold">Date of birth:</span> {{formatDate(person.dateOfBirth)}}<br />
+    {% endif %}
+
     <span class="govuk-!-font-weight-bold">CRN:</span> {{ person.crn }}</p>
   {% endif %}
 </div>
--- a/server/components/bookingListing.ts
+++ b/server/components/bookingListing.ts
@@ -6,14 +6,18 @@ import { DateFormats } from '../utils/dateUtils'
 export const personSummaryListRows = (booking: Booking): SummaryList['rows'] => {
   const rows = [
     {
-      key: textValue('Date of birth'),
-      value: textValue(DateFormats.isoDateToUIDate(booking.person.dateOfBirth)),
-    },
-    {
       key: textValue('CRN'),
       value: textValue(booking.person.crn),
     },
   ] as SummaryList['rows']
+
+  if (booking.person.dateOfBirth) {
+    rows.unshift({
+      key: textValue('Date of birth'),
+      value: textValue(DateFormats.isoDateToUIDate(booking.person.dateOfBirth)),
+    })
+  }
+
   return rows
 }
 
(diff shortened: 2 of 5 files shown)


bryangaledxw:
I think you'll also need to update `assessments/show.njk`

bryangaledxw:
If we expect a referrer to be completing a referral for someone they only have limited access to, each application page will also need to be updated. Though this could be dodged by modifying the person object when it's received in `ApplicationService.initializePage()`. It may still be worth treating the info at the top of the check your answers page as a special case though

We also rely on the name in:
`assessments/show.njk`
`applications/people/selectIndexOffence.njk`

The upcoming "select referral" page in the booking flow, which includes the name in the radio options

bryangaledxw:
I'll update the select referral radio items to handle the name not being present. This PR will conflict with #535 since we both update the pop details header (in that PR is allows the name to be a link - we'll need to be that the CRN is the link if it's a LAO). But we'll see the conflict and can deal with that when we get to it

tahb:
As I understand it and have just checked with a manual test: if a CPP puts in a CRN at the start of the referral journey that is an LAO they have no access to then they get hit with the unauthorised stick and signed out. So that's one less thing to worry about.

I could probably drop the commit that changes the personDetails component given it's only used in the referral journey right now however that might not always be true, it might be reused elsewhere. So, whilst we expect the user never to get to that component (given the LAO check on the CRN should sign them out before) we set the component up to guard in other uses.

tahb:
I've checked in with Joe to confirm that making a temporary change to the API to return a Person object that had empty fields is non-trivial. Even if it were, we'd need to guard against date of births that were not defined.

server/views/temporary-accommodation/components/pop-details-header/macro.njk
```diff
@@ -4,14 +4,20 @@
...
-      <p><span class="govuk-!-font-weight-bold">Name:</span> <a class="govuk-link" href="{{ config.nameLink }}">{{ person.name }}</a><br />
+    {% if person.name %}
+      {% if config.nameLink %}
+        <p><span class="govuk-!-font-weight-bold">Name:</span> <a class="govuk-link" href="{{ config.nameLink }}">{{ person.name }}</a><br />
+      {% else %}
+        <p><span class="govuk-!-font-weight-bold">Name:</span> {{ person.name }}<br />
+      {% endif %}
     {% else %}
-      <p><span class="govuk-!-font-weight-bold">Name:</span> {{ person.name }}<br />
+      <p><span class="govuk-!-font-weight-bold">Limited access offender<br />
```
bryangaledxw: Should this use the `nameLink` if provided?

bryangaledxw (approved):
LGTM, though I wonder if it's worth doing `{{ assessment.application.person.name | default('Limited access offender') }} in `selectAssessment.njk`, then everything manage-side is handled

server/views/temporary-accommodation/components/pop-details-header/macro.njk
tahb: Actually yes there's no reason why it shouldn't now as that page should render 👍

tahb:
[Looks like we've already guarded the assessment selection so that if there's no name, we only show the CRN](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/f213d81709f5b8bd27cdb84e098d259814efb442/server/utils/bookingUtils.ts#L383)

PR #541, opened 2023-08-09
Only enabled regions can interact with referrals journey

patinipat:
# Context

# Changes in this PR

* The referral journey is protected with by user.role and user.region
  Only referrer with `referrer` role and within an enabled region can create a referral

* Assessor will see the review referral tile when the have the correct role ``and are in an enabled region

### After
New static page /use-ndelius

image

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod


patinipat:
[Only expected regions can interact with referrals](https://trello.com/c/PBk6dQWJ/1361-only-expected-regions-can-interact-with-referrals)

server/utils/userUtils.test.ts
```diff
@@ -22,3 +25,52 @@ describe('filterProbationRegions', () => {
...
+
+  afterEach(() => {
+    config.flags.allRegionsEnabled = originalValue
+  })
+
+  it.each([
+    [true, userFactory.build({ roles: ['assessor'], region: { name: 'Kent, Surrey & Sussex' } }), true],
+    [true, userFactory.build({ roles: [], region: { name: 'Kent, Surrey & Sussex' } }), false],
+    [true, userFactory.build({ roles: ['assessor'], region: { name: 'London' } }), true],
+    [true, userFactory.build({ roles: [], region: { name: 'London' } }), false],
```
bryangaledxw: Since these are all true, can it come out of the test parameters, and into the test itself?

And I feel like it'd be more readable if the parameters were the expected true/false, the roles array, the region, and you build the user inside the test body.

Then you can write the test name as something like `'returns %p when the roles are %p and the region is %p'`

server/utils/userUtils.ts
```diff
@@ -1,6 +1,16 @@
...
-export default function filterProbationRegions(regions: ProbationRegion[], request: Request) {
+export function filterProbationRegions(regions: ProbationRegion[], request: Request) {
   return [regions.find(region => region.id === request.session.probationRegion.id)]
 }
+
+export function hasRegionEnabled(user: User): boolean {
+  return config.flags.allRegionsEnabled || !disabledRegions.includes(user.region.name)
+}
+
+export function canAssessorViewReferrals(user: User): boolean {
```
bryangaledxw: If feel like this should be `canUserViewReferrals`, as whether or not the user is an assessor is part of what this is testing

server/config.ts
```diff
@@ -52,6 +52,7 @@ export default {
   flags: {
     oasysDisabled: process.env.OASYS_DISABLED || false,
     applyDisabled: !['local', 'dev', 'test', 'preprod'].includes(environment),
+    allRegionsEnabled: ['local', 'dev', 'test'].includes(environment),
```
bryangaledxw: Can this be more descriptive? `applyEnabledForAllRegions`?

server/middleware/roleMiddleware.ts
```diff
@@ -1,14 +1,15 @@
...
+import { hasRegionEnabled } from '../utils/userUtils'
 
 export const createRoleMiddleware = (role: UserRole) => {
   return (handler: RequestHandler) => wrapHandler(handler, role)
 }
 
 const wrapHandler =
   (handler: RequestHandler, role: UserRole) => async (req: Request, res: Response, next: NextFunction) => {
-    if (res.locals.user.roles.includes(role)) {
+    if (res.locals.user.roles.includes(role) && hasRegionEnabled(res.locals.user)) {
```
bryangaledxw: I think this needs a a rethink, as AFAICT this will end up restricting the entire app to users in our enabled regions. 

Tempted to say that the signature for `wrapHandler` should be:
`const wrapHandler = (handler: RequestHandler, check: (user: User) => bool))`...

Then instead of doing:
`if (res.locals.user.roles.includes(role))`
It does:
`if (check(res.locals.user)`

And createRoleMiddleware passes in a function, but even then I think you'd need to split out our routes for this to not break things for someone.

It may be we just don't do enforcement at this level

server/@types/shared/models/TemporaryAccommodationUserRole.ts
```diff
@@ -3,4 +3,4 @@
 /* tslint:disable */
 /* eslint-disable */
 
-export type TemporaryAccommodationUserRole = 'assessor' | 'referrer';
+export type TemporaryAccommodationUserRole = 'assessor' | 'referrer' | 'cas3_referrer';
```
bryangaledxw: For my understanding, where's this change come from?

server/utils/disabledRegions.ts
```diff
@@ -0,0 +1,13 @@
+export default [
```
bryangaledxw: My instinct is it'd be safer to have an enabled list rather than a disabled list. I kinda feel like doing this without an API change to get this info is unsafe in any case, but an enabled list feels less unsafe 🙂

server/middleware/roleMiddleware.test.ts
```diff
@@ -4,26 +4,57 @@ import { userFactory } from '../testutils/factories'
...
-          roles: ['referrer'],
+          roles: ['cas3_referrer'],
+          region: { name: 'Kent, Surrey & Sussex' },
         }),
       },
     })
     const next = jest.fn()
+    config.flags.allRegionsEnabled = false
 
-    const middleware = createRoleMiddleware('referrer')
```
bryangaledxw, line 20: I feel like the change to `cas3_referrer` is worth its own commit

server/controllers/temporary-accommodation/manage/dashboardController.test.ts
```diff
@@ -1,11 +1,20 @@
 import type { NextFunction, Request, Response } from 'express'
```
bryangaledxw, line 1: In the message for this commit, "a referrals", and probably better to use "they" rather than "he"

server/utils/userUtils.ts
```diff
@@ -1,6 +1,16 @@
...
+import { ProbationRegion, TemporaryAccommodationUser as User } from '../@types/shared'
+import disabledRegions from './disabledRegions'
+import config from '../config'
 
-export default function filterProbationRegions(regions: ProbationRegion[], request: Request) {
+export function filterProbationRegions(regions: ProbationRegion[], request: Request) {
   return [regions.find(region => region.id === request.session.probationRegion.id)]
 }
+
+export function hasRegionEnabled(user: User): boolean {
```
bryangaledxw: Similar comment to above on naming. Maybe `isApplyEnabledForUser`?

bryangaledxw (commented):
Good stuff on getting your head round this. I've left a fair few comments, but I think the only fundamental thing is the change in roleMiddleware, which I think would end up breaking things for existing users.

It's the routes under `routes/apply`, and some subset (the routes to AssessmentsController) of the routes under `routes/temparary-accommodation/manage` that we'd want to be limited by region. It may be we need to split `routes/temparary-accommodation/manage` into `manage` and `place`, or we simply don't enforce this for now at the routing level

bryangaledxw:
Just realised we also have LandingController, which will need to be updated

server/config.ts
patinipat: :recycle:

server/controllers/temporary-accommodation/manage/dashboardController.test.ts
patinipat: :recycle:

server/utils/userUtils.test.ts
patinipat: :recycle:

server/utils/userUtils.ts
patinipat: :recycle:

server/middleware/roleMiddleware.ts
patinipat: Added a new middleware createUserCheckMiddleware that is defined as suggested

server/utils/disabledRegions.ts
patinipat: :recycle:  
switched to an enabledRegions list

server/middleware/roleMiddleware.test.ts
patinipat: remove cas3_referrer based on the clarification that the backend api sends referrer and not cas3_referrer

server/utils/userUtils.ts
patinipat: :recycle:

server/@types/shared/models/TemporaryAccommodationUserRole.ts
patinipat: removed this commit based on our discussion

server/views/temporary-accommodation/static/useNDelius.njk
```diff
@@ -0,0 +1,19 @@
+{% extends "partials/layout.njk" %}
+
+{% set pageTitle = "Use nDelius" %}
```
bryangaledxw: Suggest this should be:
`{% set pageTitle = applicationName + " - Referrals should be submitted through nDelius" %}`
to match conventions elsewhere

server/views/temporary-accommodation/static/useNDelius.njk
```diff
@@ -0,0 +1,19 @@
...
+
+{% block content %}
+  <div class="govuk-grid-row">
```
bryangaledxw, line 6: I think in the GDS system you want rows, columns, then content, so this should contain something like:
`<div class="govuk-grid-column-two-thirds">`

server/utils/userUtils.ts
```diff
@@ -1,6 +1,28 @@
...
-export default function filterProbationRegions(regions: ProbationRegion[], request: Request) {
+export function filterProbationRegions(regions: ProbationRegion[], request: Request) {
   return [regions.find(region => region.id === request.session.probationRegion.id)]
 }
+
+export function isApplyEnabledForUser(user: User): boolean {
+  return config.flags.applyEnabledForAllRegions || enabledRegions.includes(user.region.name)
+}
+
+export function canUserViewReferrals(user: User): boolean {
```
bryangaledxw: Sorry, gonna be annoying about naming 😅 - think change to `userHasAssessorRoleAndIsApplyEnabled`, so we're staying consistent within `userUtils`

patinipat: :recycle:

server/views/temporary-accommodation/static/useNDelius.njk
patinipat: :recycle:

server/middleware/userCheckMiddleware.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    const middleware = createUserCheckMiddleware(success)
+
+    middleware(handler)(request, response, next)
+
+    expect(handler).toHaveBeenCalledWith(request, response, next)
+  })
+
+  it('returns middleware that calls next() when the user predicate fails and no redirectPath provided', async () => {
```
bryangaledxw: Think the reference to `redirectPath` is a leftover

server/routes/apply.ts
```diff
@@ -3,14 +3,19 @@
 import type { Router } from 'express'
 import type { Controllers } from '../controllers'
 import Apply from '../form-pages/apply'
-import { createRoleMiddleware } from '../middleware/roleMiddleware'
+import { createUserCheckMiddleware } from '../middleware/userCheckMiddleware'
```
bryangaledxw, line 6: I think go ahead and delete roleMiddleware at this step

server/controllers/temporary-accommodation/manage/dashboardController.ts
```diff
@@ -1,10 +1,10 @@
 import type { Request, RequestHandler, Response } from 'express'
-import { canUserViewReferrals } from '../../../utils/userUtils'
+import { userHasAssessorRoleAndIsApplyEnabled } from '../../../utils/userUtils'
```
bryangaledxw: I think just have it as `userHasAssessorRoleAndIsApplyEnabled` from the first commit, rather than add the function and rename it later. dxw conventions are to structure your commits as though you knew the end result from the start, so it's telling an idealised story of how a feature was built. Definitely takes a bit of getting used to, but it keeps commit history clean and it makes a lot of difference when it comes to reviewing PRs

It might be worth dropping this commit entirely, and rolling its changes into the first two commits

server/routes/apply.ts
```diff
@@ -5,7 +5,6 @@ import type { Controllers } from '../controllers'
 import Apply from '../form-pages/apply'
 import { createUserCheckMiddleware } from '../middleware/userCheckMiddleware'
 import paths from '../paths/apply'
-import staticPaths from '../paths/temporary-accommodation/static'
```
bryangaledxw: Think this belongs in the previous commit

bryangaledxw (commented):
Looks good. In terms of code changes it's pretty there. Think it's worth doing a sweep over the commit messages - I spotted "will protecting" in the first commit - and the commit tidying I've suggested. I'll give this some manual testing, as this is a risky one, but if you can do these last few changes, and fixing that test failure doesn't need any big changes, this should be good to go in.

server/routes/temporary-accommodation/index.ts
```diff
@@ -22,10 +21,7 @@ export default function routes(controllers: Controllers, services: Services): Ro
   get(paths.static.useNDelius.pattern, staticController.useNDelius(), { auditEvent: 'VIEW_USE_NDELIUS' })
 
   manageRoutes(controllers, services, router)
-
-  if (!config.flags.applyDisabled) {
```
bryangaledxw, line 25: Think remove this flag from `config.ts` entirely at this point, since it's no longer used

server/views/temporary-accommodation/static/useNDelius.njk
patinipat: used `<div class="govuk-grid-column-full">`

server/routes/temporary-accommodation/index.ts
patinipat: :recycle:

server/routes/apply.ts
patinipat: :recycle:

server/controllers/temporary-accommodation/manage/dashboardController.ts
patinipat: :recycle:

server/routes/apply.ts
patinipat: :+1:

server/middleware/userCheckMiddleware.test.ts
patinipat: good spot

server/controllers/temporary-accommodation/manage/bookingsController.test.ts
```diff
@@ -30,12 +30,21 @@ import {
   insertBespokeError,
   insertGenericError,
 } from '../../../utils/validation'
+import { isApplyEnabledForUser } from '../../../utils/userUtils'
```
bryangaledxw, line 32: Think swap this commit with the previous one, so we're removing the dependence on `applyDisabled` being in `nunjucksSetup` before we remove `applyDisabled` from `nunjucksSetup`

server/controllers/temporary-accommodation/manage/bookingsController.test.ts
```diff
@@ -30,12 +29,21 @@ import {
...
 jest.mock('../../../utils/restUtils')
 jest.mock('../../../utils/validation')
 jest.mock('../../../utils/utils')
+jest.mock('../../../utils/userUtils', () => {
+  const module = jest.requireActual('../../../utils/userUtils')
+
+  return {
+    ...module,
+    isApplyEnabledForUser: jest.fn(),
+  }
```
bryangaledxw: I wouldn't worry about this tbh, think just jest.mock('../../../utils/userUtils') is fine

There's a few odd cases in the codebase where we need a mocked version of a function and a real version in the same test file, but those are the only places we really use this pattern

bryangaledxw (approved):
Couple of last bits of feedback, but otherwise good for this to go in

On the merge commit at the end - we generally rebase on top of main, rather than doing merge commits, to keep the tree linear. Not sure if that'll make any difference once it's merged in though

server/controllers/temporary-accommodation/manage/bookingsController.test.ts
patinipat: :recycle:

server/controllers/temporary-accommodation/manage/bookingsController.test.ts
patinipat: :recycle:

PR #559, opened 2023-08-14
Add human notes to a referral

patinipat:
# Context

# Changes in this PR

Add a create assessment note form on the review and assess show page
Notes are displayed using the moj Timeline component

### Before
localhost_3000_review-and-assess_df0e4635-b71d-4ad2-8191-d74b1bcb27ad_df0e4635-b71d-4ad2-8191-d74b1bcb27ad_summary_success=false

### After
localhost_3000_review-and-assess_df0e4635-b71d-4ad2-8191-d74b1bcb27ad_df0e4635-b71d-4ad2-8191-d74b1bcb27ad_summary_success=true

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/server/utils/nunjucksSetup.ts
+++ b/server/utils/nunjucksSetup.ts
@@ -19,6 +19,7 @@ import { statusTag as personStatusTag } from './personUtils'
 import { initialiseName, mapApiPersonRisksForUi, removeBlankSummaryListItems, sentenceCase } from './utils'
 
 import { dashboardTableRows, taskResponsesToSummaryListRowItems } from './applicationUtils'
+import * as AssessmentUtils from './assessmentUtils'
 import * as AttachDocumentsUtils from './attachDocumentsUtils'
 import * as BedspaceSearchResultUtils from './bedspaceSearchResultUtils'
 import * as OasysImportUtils from './oasysImportUtils'
@@ -179,6 +180,7 @@ export default function nunjucksSetup(app: express.Express, path: pathModule.Pla
   njkEnv.addGlobal('TasklistUtils', TasklistUtils)
   njkEnv.addGlobal('PremisesUtils', PremisesUtils)
   njkEnv.addGlobal('OasysImportUtils', OasysImportUtils)
+  njkEnv.addGlobal('AssessmentUtils', AssessmentUtils)
   njkEnv.addGlobal('AttachDocumentsUtils', AttachDocumentsUtils)
   njkEnv.addGlobal('PhaseBannerUtils', PhaseBannerUtils)
 }
--- a/server/utils/nunjucksSetup.ts
+++ b/server/utils/nunjucksSetup.ts
@@ -1,11 +1,12 @@
 /* eslint-disable no-param-reassign */
 /* istanbul ignore file */
 
-import * as pathModule from 'path'
 import express from 'express'
 import nunjucks from 'nunjucks'
+import * as pathModule from 'path'
 
 import type { ErrorMessages, PersonStatus } from '@approved-premises/ui'
+import { statusTag as assessmentStatusTag } from './assessmentUtils'
 import { DateFormats, dateInputHint } from './dateUtils'
 import {
   convertObjectsToCheckboxItems,
@@ -15,7 +16,6 @@ import {
   parseNaturalNumber,
 } from './formUtils'
 import { statusTag as personStatusTag } from './personUtils'
-import { statusTag as assessmentStatusTag } from './assessmentUtils'
 import { initialiseName, mapApiPersonRisksForUi, removeBlankSummaryListItems, sentenceCase } from './utils'
 
 import { dashboardTableRows, taskResponsesToSummaryListRowItems } from './applicationUtils'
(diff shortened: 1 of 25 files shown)


bryangaledxw:
Giving this a preliminary pass, all looks good. Happy with the commit structure, building up the layers. I think maybe it'd be worth putting the new util method (and its inclusion in nunjucks setup) in it's own commit before the controller changes, but that's a minor thing. Some other things I spotted:

* Should `text: note.message` in assessment utils be `html: formatLines(note.message)`? Just want to make sure notes with line breaks are handled properly. Suggest it's worth using `faker.lorem.lines()` in the factory to exercise this
* Do you need any error handling in the new controller method. Does/should the API allow empty notes? You'd need to do the same style of error handling as in `PremisesController.create()` and add a new message to `errors.json`

libuk (commented):
Looks good. I agree with @bryangaledxw that the utils change could be pulled out into a separate commit, but shouldn't be a blocker. I'll defer to him for approval as there may be some context I'm missing.

patinipat:
Extract the new helper function in assessmentUtils in its own commit

server/views/temporary-accommodation/assessments/show.njk
```diff
@@ -66,6 +69,38 @@
...
+        {{
+        govukTextarea({
+          name: "note",
+          id: "note",
+          label: {
+            text: "Add a note",
+            classes: "govuk-label--m referral-note"
+          }
+        })
+        }}
```
bryangaledxw: Suggest use taTextarea here instead, it'll save some repetition. And use the same naming as the field in the API. If this is "message" rather than "notes", everything will link up by magic (ie. `catchValidationErrorOrPropogate`) if there's an error from the API - this'll be be highlighted in red, and the errors at the top of the page will link to here

server/utils/nunjucksSetup.ts
```diff
@@ -92,6 +92,7 @@ export default function nunjucksSetup(app: express.Express, path: pathModule.Pla
   njkEnv.addGlobal('dateFieldValues', dateFieldValues)
   njkEnv.addGlobal('formatDate', DateFormats.isoDateToUIDate)
   njkEnv.addGlobal('formatDateTime', DateFormats.isoDateTimeToUIDateTime)
+  njkEnv.addFilter('mojDate', DateFormats.isoDateTimeToUIDateTime)
```
bryangaledxw: So is this used internally by the timeline component?

cypress_shared/pages/assess/show.ts
```diff
@@ -52,4 +67,36 @@ export default class AssessmentShowPage extends Page {
...
+      })
+  }
+
+  createNote(message: string) {
+    cy.get('.referral-note')
+      .contains('Add a note')
+      .parent()
+      .parent()
+      .within(() => {
+        cy.get('.govuk-textarea').type(message)
```
bryangaledxw: Think this could be a call down to `completeTextInputByLabel` in the base class - `completeTextInputByLabel('Add a note', message)`. And getting more into personal preference, consider just doing something like `cy.get('button').contains('Save note').click()`

bryangaledxw (commented):
Gone through a bit more closely. I'll hold off regarding as done until you've had a chat with Nick about showing the name of the person who created the note, but this doesn't look far off, good stuff

server/views/temporary-accommodation/assessments/show.njk
patinipat: :recycle:

cypress_shared/pages/assess/show.ts
patinipat: :recycle:

server/utils/nunjucksSetup.ts
patinipat: yes https://github.com/ministryofjustice/moj-frontend/blob/main/src/moj/components/timeline/template.njk#L18

patinipat:
Added support to display the name of the staff member who created the note. This relies on the `User` type being updated to provide the `name` property. This commit https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/559/commits/678d2299e68e504bd32da39f36ebd13ac3bec44f is there in case we need to format the user.name has Im not sure what will format will be provided.

bryangaledxw:
@libuk Ready for another look

wiremock/assessmentStubs.ts
```diff
@@ -86,6 +87,19 @@ const assessmentStubs = [
             jsonBody: {},
           },
         },
+        {
+          priority: 99,
```
libuk, line 91: What does this priority key do out of interest?

libuk (approved):
Nice work 👍🏾

wiremock/assessmentStubs.ts
bryangaledxw: Generally when we're stubbing a POST endpoint, we'll have the success response that matches against a path, and some error responses that match against the path plus specific fields being missing. We want to make sure that the success response takes lower priority, otherwise that'd match first and we'd never see the error response in our wiremock API

Worth having a look at `errorStub()` and `getCombinations()`. We don't use the latter here as there's only one field (`message`), but generally we use the two together to generate error stubs for any combination of missing fields

PR #570, opened 2023-08-21
Introduce assessment summary page

libuk:
# Context

# Changes in this PR
* Introduces assessment summary page

### Before
Screenshot 2023-08-23 at 12 13 37

### After
Screenshot 2023-08-23 at 12 09 05
Screenshot 2023-08-23 at 14 09 28

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod


libuk:
[Users can see an application summary page](https://trello.com/c/4M8t7nx2/1322-users-can-see-an-application-summary-page)

libuk:
@bryangaledxw This is ready for review

bryangaledxw (approved):
LGTM

Only thing that's bugging me slightly (and not a blocker if there's not a good solution) is that the breadcrumbs don't strictly "grow" as you progress down them, going from "Referrals > Summary - Name" to "Referrals > Summary > Name"

I wonder if it'd make sense for the full page to have a back link instead, thinking of it as a sideways offshoot rather than a decent down a hierarchy?

I feel like it's worth giving some attention to the page title for these two pages at the same time - can the page title be consistent with the top breadcrumb?

libuk:
> I wonder if it'd make sense for the full page to have a back link instead, thinking of it as a sideways offshoot rather than a decent down a hierarchy?

Agreed, I think this is an elegant solution.

> I feel like it's worth giving some attention to the page title for these two pages at the same time - can the page title be consistent with the top breadcrumb?

Good spot will fix it. I think going with the following makes sense if we think of the full referral as an offshoot of sorts:

Summary: `Temporary Accommodation - Aadland Bertrand referral summary`
Full: `Temporary Accommodation - Aadland Bertrand referral`

It's not perfect but now that I think about it, it might be a good idea to create a task for looking at the page title across the site. Other gov websites tend to have the page name first. I'll make this change and then write up a [card](https://trello.com/c/vfPm0mIb/1502-audit-of-page-titles-across-the-app).

PR #578, opened 2023-08-24
Handle empty turnaround in booking info component

bryangaledxw:
We update our booking info component to handle the turnaround field on a booking being undefined. We already handle this possibility in other places we work with turnarounds, but missed this instance

### DIFF
--- a/server/components/bookingInfo.test.ts
+++ b/server/components/bookingInfo.test.ts
@@ -552,5 +552,27 @@ describe('BookingInfo', () => {
         ]),
       )
     })
+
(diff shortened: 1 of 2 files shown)


bryangaledxw:
@tahb Found this by accident while checking LAOs on dev - going by the logs, this was a culprit for me not being able to view one of the bookings visible on my personal dev log in. Wonder if it's related to the crashing booking Jack mentioned

Error was `TypeError: Cannot read properties of null (reading 'workingDays')`, and this seemed the only plausible source

tahb (approved):
LGTM. Will draw up a card to see if the API can prevent this.

PR #580, opened 2023-08-24
Update assessment links

bryangaledxw:
We update the back link on the state change confirmation page, and the link on the booking page to link to the assessment summary, rather than the full assessment


libuk (approved):
Good catch

PR #583, opened 2023-08-25
Fix select offence referral page

bryangaledxw:
We update offenceTableRows() to handle the offence date possibly being empty

localhost_3000_referrals_people_C954470_select-offence


tahb (approved):
Super, thanks for jumping on this.

PR #586, opened 2023-08-29
Hide system notes from assessments

bryangaledxw:
An [upcoming API PR](https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/919) will add system notes to assessments. To prevent these showing in a way that will be confusing to users, we hide them until we can implement display of system notes properly.

Though the upcoming PR will send a 'type' field with each note, we cannot use that yet, and to resolve this chicken-and-egg situation, we check if 'message' is populated to determine if this is a system note.

### DIFF
--- a/server/utils/assessmentUtils.test.ts
+++ b/server/utils/assessmentUtils.test.ts
@@ -236,10 +236,9 @@ describe('assessmentUtils', () => {
         createdAt: '2024-05-01',
       })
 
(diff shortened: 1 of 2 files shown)


bryangaledxw:
[Check frontend against system notes API update](https://trello.com/c/z6bGJU2U/1505-check-frontend-against-system-notes-api-update)

libuk (approved):

PR #656, opened 2023-09-28
Show offence ID on check your answers and full referral pages

libuk:
# Context
HPTs currently have no way of knowing which index offence a referral belongs to and are having to ask the CPP, therefore, we want to show the offence ID on the referral page under the 'sentence information' section.

To achieve this we are adding the offence ID to the response of the first page of the 'sentence information' task. As there is no way to change the offence ID without starting a new referral, we hide the "change" action on the check your answers page.

This isn't an ideal solution as it doesn't follow the conventions of the other form pages, but here is some rationale:

* Offence ID is fetched by the server without the need for user input. We only ask the user to choose an offence ID if more than one offence is retrieved from the API. This means making a TaskListPage for offence ID non-trivial and would require a different approach to what we have now. Given there are limited dev days left on CAS3, taking on this work isn't feasible.

* I thought about adding a new type of TaskListPage which wouldn't be visible to the user, but changing the existing domain model for this one instance felt a bit premature, and would still be a good chunk of work.

* The solution in this commit is a hack but is fairly non-invasive and easy to change in the future.

Let me know if this is something you've encountered in CAS1/CAS2. Alternative solutions are welcome!

# Changes in this PR

### Before
referral_before
check_your_answers_before

### After
referral_after
check_your_answers_after

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/server/form-pages/apply/accommodation-need/sentence-information/offendingSummary.test.ts
+++ b/server/form-pages/apply/accommodation-need/sentence-information/offendingSummary.test.ts
@@ -47,6 +47,7 @@ describe('OffendingSummary', () => {
       const page = new OffendingSummary(body, application)
 
       expect(page.response()).toEqual({
+        'Offence ID': application.offenceId,
         'Summary of offending history': 'Offending summary',
       })
     })
--- a/server/form-pages/apply/accommodation-need/sentence-information/offendingSummary.ts
+++ b/server/form-pages/apply/accommodation-need/sentence-information/offendingSummary.ts
@@ -9,6 +9,8 @@ export type OffendingSummaryBody = {
   summary: string
 }
 
+export const offenceIdKey = 'Offence ID'
+
 @Page({ name: 'offending-summary', bodyProperties: ['summary'] })
 export default class OffendingSummary implements TasklistPage {
   title: string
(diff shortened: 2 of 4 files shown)


libuk:
[Referrals not storing the index offence](https://trello.com/c/JcaSOG7i/1595-referrals-not-storing-the-index-offence)

richpjames (approved):
This approach makes sense to me given the time constraints. 

In CAS1 we handle index offences as the first page of the Apply journey. Behind the scenes an application is not created until an index is selected (if there is only one offence it is selected by default). This means that the index offence is dealt with before we hit the pages controller and has its own controller method. This approach obviously isn't feasible for CAS3 as it would take a long time to put into place. 

My one criticism would be that it doesn't look like the new functionality in the `checkYourAnswers` utils file is unit tested.

selina-dxw:
Looks good, thanks!

libuk:
> My one criticism would be that it doesn't look like the new functionality in the checkYourAnswers utils file is unit-tested.

Good spot, thanks @richpjames

PR #662, opened 2023-10-03
Remove check for in progress applications in e2e test

libuk:
# Context
After submitting an application we were attempting to check the list of in-progress applications in our e2e tests.

However, as the list of in-progress applications was never provided to `ListPage`, we would always expect the length to be 0. This caused failures on CI when there were in-progress applications present.

Instead of fixing this test, I'm choosing to remove it as we already check for the confirmation page beforehand. Also, the completed application would never show up as in progress after submission, so this check isn't providing any value.

image

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod


tahb (approved):
I think this makes sense. When we add the new feature for viewing submitted referrals then we can have the test coverage in there.

PR #681, opened 2023-10-13
Feature/support new service url in dev

tahb:
# Context

# Changes in this PR

To support the new URL alongside the old I've done some name refactoring to `domain` and `INGRESS_URL`. This hopefully should not change any behaviour.

We then add the new configuration so we can support a second domain. It's at this point I'm unsure if we need to add an [additional passport strategy](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/87416339c2fdec6ad4e150b1ee63b31dcaac7dac/server/authentication/auth.ts#L38C54-L38C54) or if it'll be able to redirect the user between domains? Such as if a user arrives via `transitional.x` then they can sign in and be taken to `temporary.x`. If it works in this direction then I imagine that we can eventually reverse it so that anyone signing in on the older temporary.x will be redirected to the `transitional.x` callback url.

We then enable the new domain in the dev environment only as a test. If we wanted to be more cautious and not risk blocking the pipeline we could move this over to the test environment which is better isolated as it's not used for e2e tests?

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -13,7 +13,7 @@ generic-service:
   env:
     ENVIRONMENT: dev
     APPROVED_PREMISES_API_URL: 'https://approved-premises-api-dev.hmpps.service.justice.gov.uk'
-    INGRESS_URL: 'https://temporary-accommodation-dev.hmpps.service.justice.gov.uk'
+    FIRST_INGRESS_URL: 'https://temporary-accommodation-dev.hmpps.service.justice.gov.uk'
     HMPPS_AUTH_URL: 'https://sign-in-dev.hmpps.service.justice.gov.uk/auth'
     TOKEN_VERIFICATION_API_URL: 'https://token-verification-api-dev.prison.service.justice.gov.uk'
     AUDIT_LOG_ERRORS: 'true'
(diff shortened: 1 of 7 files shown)


libuk (commented):
> If we wanted to be more cautious and not risk blocking the pipeline we could move this over to the test environment which is better isolated as it's not used for e2e tests?

I'd say it's a good shout to only turn on in the test environment for the time being.

> We then add the new configuration so we can support a second domain. It's at this point I'm unsure if we need to add an [additional passport strategy](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/87416339c2fdec6ad4e150b1ee63b31dcaac7dac/server/authentication/auth.ts#L38C54-L38C54) or if it'll be able to redirect the user between domains?

I'm also not sure if we'd need a new strategy, so the changes made in this PR makes sense for initial test IMO 👍🏾

tahb:
@libuk Thanks for reviewing. I've switched that over to start on the last commit to the test environment 👍

libuk (approved):
🎉

PR #688, opened 2023-10-17
Revert "Enable transitional domain in dev"

tahb:
Reverts ministryofjustice/hmpps-temporary-accommodation-ui#687


libuk (approved):

tahb:
I think we'll need this anyway and having it doesn't seem to affect sign in on dev.

PR #691, opened 2023-10-18
Redirect requests from the old domain to the new one

tahb:
# Context

# Changes in this PR

[The previous attempt to allow users to sign into the new domain name](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/681) failed and had to be [reverted](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/686).

This is probably due to the fact we were using two strategies with the same callback url. It always redirected the user to transitional-accommodation.x despite starting from temporary. This appeared worked when we tested locally but on a real environment we didn't see the same thing. Presumably a problem in the fiddly attempt to replicate locally.

This time we change approach.

We are going to try first redirecting all traffic from the old url to the new one. From there we can go back to a single passport strategy. That way when the user signs in they're always signing in on the "transitional" equivalent.

In order to protect production while we test this sensitive code change we add some plumbing in so that the new behaviour only works in the test environment for now.

What i'd like to observe in test is that I arrive at temporary-accommodation and am redirected to transitional-accommodation. From there I can sign in and am taken back to transitional-accommodation successfully.

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/server/middleware/setUpDomainRedirect.test.ts
+++ b/server/middleware/setUpDomainRedirect.test.ts
@@ -1,0 +1,46 @@
+import express, { Express } from 'express'
+import request from 'supertest'
+import setUpDomainRedirect from './setUpDomainRedirect'
+import config from '../config'
+
+jest.mock('../config', () => ({
+  secondDomain: 'http://example.com',
+  firstDomain: 'http://example.org',
+}))
+
+const setupApp = (): Express => {
+  const app = express()
+  app.use(setUpDomainRedirect())
+
+  app.get('/known', (_req, res, _next) => {
+    res.send('known')
+  })
+
+  return app
+}
+describe('setUpDomainRedirect', () => {
+  it('should redirect to the target domain when in test mode', () => {
+    process.env.NODE_ENV = 'test'
+    const app = setupApp()
+    const targetHost = new URL(config.firstDomain).host
+    return request(app).get('/known').set('host', targetHost).expect(301)
+  })
+
+  it('should not redirect requests already coming from the second domain', () => {
+    process.env.NODE_ENV = 'test'
+    const app = setupApp()
+    const targetHost = new URL(config.secondDomain).host
+    return request(app).get('/known').set('host', targetHost).expect(200)
+  })
+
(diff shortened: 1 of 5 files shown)


server/middleware/setUpDomainRedirect.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  })
+
+  return app
+}
+describe('setUpDomainRedirect', () => {
+  it('should redirect to the target domain when in test mode', () => {
+    process.env.NODE_ENV = 'test'
+    const app = setupApp()
+    const targetHost = new URL(config.firstDomain).host
+    return request(app).get('/known').set('host', targetHost).expect(301)
```
libuk, line 26: [suggestion] I know this was copied from the other test, but I don't think we need the `return` statement in these tests.

libuk (approved):
Looks good, just one comment which is by no means a blocker, especially as we're just testing.

server/middleware/setUpDomainRedirect.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  })
+
+  it('should not redirect requests already coming from the second domain', () => {
+    process.env.NODE_ENV = 'test'
+    const app = setupApp()
+    const targetHost = new URL(config.secondDomain).host
+    return request(app).get('/known').set('host', targetHost).expect(200)
```
libuk, line 33: Your current implementation is in line with how we do it elsewhere in the codebase, so happy for this to stay as is.

Looking at the examples from [the docs](https://github.com/ladjs/supertest), we can avoid using `return` to ensure the promise is resolved if we use `done()` `.end()` or `async/await`:

```ts
it('should not redirect requests already coming from the second domain', async () => {
    process.env.NODE_ENV = 'test'
    const app = setupApp()
    const targetHost = new URL(config.secondDomain).host
    const response = await request(app).get('/known').set('host', targetHost)
    expect(response.status).toEqual(200)
})
```

PR #699, opened 2023-10-19
Redirect requests from the old domain to the new one

tahb:
Reverts ministryofjustice/hmpps-temporary-accommodation-ui#698 plus adds a correction to how we check the current environment to use [config.environment](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/b1ae840baa930ce17373d2fa2616fcc7c244a8f5/server/config.ts#L45) which I was lucky to catch after investigating the e2e errors we now believe are being caused by the api.

### DIFF
--- a/server/authentication/auth.ts
+++ b/server/authentication/auth.ts
@@ -29,15 +29,27 @@ const authenticationMiddleware: AuthenticationMiddleware = verifyToken => {
 }
 
 function init(): void {
+  const defaultStrategyConfig = {
+    authorizationURL: `${config.apis.hmppsAuth.externalUrl}/oauth/authorize`,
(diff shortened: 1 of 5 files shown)


server/authentication/auth.ts
```diff
@@ -40,7 +40,7 @@ function init(): void {
 
   // TODO: Remove this once our sign in testing on our new domain is complete on
   // the test env.
-  const redirectToDomain = process.env.NODE_ENV === 'test' ? config.secondDomain : config.firstDomain
```
libuk: Good catch

libuk (approved):

PR #732, opened 2023-11-09
Migrating project to use groups in allowlist

andrewrlee:
This PR migrates the project to use groups of IPs in their allowlist.

By referring to groups to IP addresses, we can centralize the definition of groups of ip addresses.
If these lists require changing in the future, we can change the definition once and future deploys across all services will automatically include these new IPs.

1 allowlist(s) have been detected that can be migrated.

### New Groups

The effect of applying this PR is as follows:

- The following groups will be applied: `internal,prisons`
- The size of the allowlist defined in this file will change: `15 => 0 (15 removed)`

### Added IPs

The new Group membership will result in the following IPs being added to your allowlist by applying this PR:

- moj-official-tgw-prod
  

- moj-official-tgw-preprod
  

- moj-official-ark-c-vodafone
  

- moj-official-ark-f-vodafone
  

- mojo-azure-landing-zone-public-egress-1
  

- mojo-azure-landing-zone-public-egress-2
  

- mojo-azure-landing-zone-public-egress-3
  

- mojo-azure-landing-zone-public-egress-4
  

### Removed IPs

The following IPs have been identified as unnecessary and will be removed by applying this PR:

- health-kick

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  hmpps: ministryofjustice/hmpps@7.3.3
+  hmpps: ministryofjustice/hmpps@7
   slack: circleci/slack@4.10.1
   node: circleci/node@4.5.2
 
--- a/helm_deploy/hmpps-temporary-accommodation-ui/Chart.yaml
+++ b/helm_deploy/hmpps-temporary-accommodation-ui/Chart.yaml
@@ -5,7 +5,7 @@ name: hmpps-temporary-accommodation-ui
 version: 0.2.0
 dependencies:
   - name: generic-service
-    version: 2.8.1
+    version: "2.8"
     repository: https://ministryofjustice.github.io/hmpps-helm-charts
   - name: generic-prometheus-alerts
     version: 1.3.3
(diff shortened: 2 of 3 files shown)


helm_deploy/hmpps-temporary-accommodation-ui/Chart.yaml
```diff
@@ -5,7 +5,7 @@ name: hmpps-temporary-accommodation-ui
 version: 0.2.0
 dependencies:
   - name: generic-service
-    version: 2.8.1
+    version: "2.8"
```
tahb, line 8: What's the rational for removing the patch and going backwards? Is this not a downgrade?

.circleci/config.yml
```diff
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  hmpps: ministryofjustice/hmpps@7.3.3
+  hmpps: ministryofjustice/hmpps@7
```
tahb, line 4: Again, what's the rationale with this one? What breaks without this change? Is setting only the major here not the equivalent of a downgrade? [It looks like the latest version number of 7 is 7.5.1](https://circleci.com/developer/orbs/orb/ministryofjustice/hmpps).

tahb (changes requested):
Hey Andrew, thanks for this, we definitely need to proactively allow the right IP ranges 👍 

Some questions around version changes.

And in general, has this changeover been tested by any other service yet? If not, I think we should consider trying this on preprod first and testing it out.

.circleci/config.yml
andrewrlee: btw- this wasn't a downgrade, we recommend floating on the major version of the orb and helm chart

froddd:
Superseded by #983

PR #764, opened 2023-11-23
Pagination and Table Sorting

andrewcharnley:
This feature commit adds pagination and server based table sorting.

It also adds a minimum number of bookings to allow this feature to be e2e tested.

Screenshot 2023-11-23 at 16 49 51

# Context

[](https://trello.com/c/vhrk6bcy/1684-booking-search-timing-out-when-searching-arrived-and-departed-bookings)

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod


cypress.config.e2e.ts
```diff
@@ -18,6 +18,14 @@ async function setupNodeEvents(
     }),
   )
 
+  on('task', {
+    log(...args) {
+      // eslint-disable-next-line
+      console.log(...args)
```
libuk, line 24: [question] Is this for better visibility when running tests?

e2e/tests/stepDefinitions/manage/bookingSearch.ts
```diff
@@ -55,3 +55,90 @@ Then('I should see a summary of the booking on the departed bookings page', () =
...
+    cy.get('a.govuk-pagination__link').eq(1).click()
+    cy.url({ timeout: 10000 }).should('include', '&page=2')
+    cy.get('.govuk-pagination__item').eq(1).should('have.class', 'govuk-pagination__item--current')
+  })
+})
+
+Then('I should see different results', () => {
+  cy.then(function _() {
+    cy.get('tbody >tr').should('exist')
+    cy.get('tbody >tr').invoke('html').should('not.equal', this.results)
```
libuk, line 84: Nice!

libuk (approved):
Looks good. As I mentioned, I find it easier to review big changes like this in smaller linear commits. But this is a larger code standards concern which I think should be addressed outside the context of a single PR.

tahb:
Thanks for reviewing @libuk. To echo your point, we should all be aiming for smaller atomic commits that tell the story as [documented in our playbook](https://playbook.dxw.com/tech/how-we-do-development/#5-commit-changes-in-small-steps). One of the benefits of investing more energy into it is that it makes reviewing easier and more valuable, especially when the reviewer has much less context than the author. Let's please invest more into it in future pull requests.

cypress.config.e2e.ts
andrewcharnley: Yep, it's otherwise a nightmare to find where data is tripping up in headless.

Usage cy.task('log', something)

andrewcharnley:
> Thanks for reviewing @libuk. To echo your point, we should all be aiming for smaller atomic commits that tell the story as [documented in our playbook](https://playbook.dxw.com/tech/how-we-do-development/#5-commit-changes-in-small-steps). One of the benefits of investing more energy into it is that it makes reviewing easier and more valuable, especially when the reviewer has much less context than the author. Let's please invest more into it in future pull requests.

While I agree to follow the dxc standard I disagree with the reasoning behind it. Development within commits is rarely linear, they may be reverted or entirely rewritten, none of which seems helpful to the reviewer. Case in point, Dan had created three commits for this feature and I had to change what they did as more knowledge surfaced. The other aspect is it makes reverting somewhat more difficult.

andrewcharnley:
> Looks good. As I mentioned, I find it easier to review big changes like this in smaller linear commits. But this is a larger code standards concern which I think should be addressed outside the context of a single PR.

Perhaps it's the lack of in-line code documentation around reasoning, and smaller commits trying to make up for it?

e2e/tests/stepDefinitions/manage/bookingSearch.ts
andrewcharnley: Took me ages to find that .invoke function. Jquery awfulness!

tahb:
Indeed development isn't linear. We can however use git often to manipulate the work as we go to curate the commits. For that example it's normal for those 3 commits to not end up having everything they needed and for them be edited or fully dismantled and represented when we learn more.

Reverting is straightforward when we couple this approach with merge commits. We're then able to revert the sum by reverting the merge commit or any individual part/s if the rest is good.

We consider inline comments as a signal that the code is not presenting itself clearly enough. Any context or additional considerations that we often have can be documented and found in the authoring commit.

PR #684, opened 2023-10-16
Feature/1588 add reporter role

libuk:
# Context
We want to add a new 'reporter' role which will enable reporters to
download reports for all regions.

# Changes in this PR
* Add `reporter` role
* Allow reporters to only see `reports/bookings` page
* Allow reporters to download reports from all regions

### Before
Screenshot 2023-10-18 at 15 28 36

### After
Screenshot 2023-10-18 at 15 22 40

### Screen recording

https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/assets/6377078/cec89a99-7377-4a14-b7b0-9f7fdee89c48

# Release checklist

[Release process documentation](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/edit-v2/4247847062?draftShareId=a1c360ab-bd31-4db1-aae3-7cc002761de9).

## Pre-merge checklist

- [ ] Are any changes required to dependent services for this change to work?
  (eg. CAS API)
- [ ] Have they been released to production already?

## Post-merge checklist

- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to test
- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to preprod
- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to prod

PR #765, opened 2023-11-27
UI Context switching for Premises Index/List Display.

andrewcharnley:
A user with region of Greater Manchester will see properties with Local Authority area.

As the API doesn't provide UI flags a local config is held and augments the data it provides. Doing so requires extending the shared types.

Screenshot 2023-11-27 at 13 30 05

Screenshot 2023-11-27 at 13 39 09

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod


libuk (commented):
I understand the reasoning behind the approach, but I'm not sure the API should be sending a UI flag in this instance.

The introduction of `CallConfigLocal` indicates that the logic is leaking out into different areas of the codebase which I find harder to reason about.

If we added an explicit check for the user's region ID, we could avoid the extra overhead and keep the logic local to where it's needed.

So in our premises controller, this

```ts
useLocalAuthorityTableHeader: callConfig.probationRegion?.config?.flags.properties.useLAnotPDU
```

Could potentially be written as

```ts
const manchesterProbationRegionId = 'manchester-id'

// ...

useLocalAuthorityTableHeader: req.user.region.id === manchesterProbationRegionId
```

It's crude, but it's clear what's happening at a glance. If more instances like this start to come up in future, we could introduce a more formalised approach, such as enums and/or helpers.

andrewcharnley:
> I understand the reasoning behind the approach, but I'm not sure the API should be sending a UI flag in this instance.
> 
> The introduction of `CallConfigLocal` indicates that the logic is leaking out into different areas of the codebase which I find harder to reason about.
> 
> If we added an explicit check for the user's region ID, we could avoid the extra overhead and keep the logic local to where it's needed.
> 
> So in our premises controller, this
> 
> ```ts
> useLocalAuthorityTableHeader: callConfig.probationRegion?.config?.flags.properties.useLAnotPDU
> ```
> 
> Could potentially be written as
> 
> ```ts
> const manchesterProbationRegionId = 'manchester-id'
> 
> // ...
> 
> useLocalAuthorityTableHeader: req.user.region.id === manchesterProbationRegionId
> ```
> 
> It's crude, but it's clear what's happening at a glance. If more instances like this start to come up in future, we could introduce a more formalised approach, such as enums and/or helpers.

I wouldn't say it's just local to this function, it's a properly of users region. There may be other areas where the  user requires displaying LA instead of PDU. I wonder if the 'booking overview' widget I have seen a few times is one?

I think it's better to at least extract

 const manchesterProbationRegionId = 'manchester-id'
 
 Into config, which now lives at /etc/config.json. I'm adverse to any 'config in code' practice.
 
 But ultimately I agree, it's a property of a user's region and the single source of truth should provide it.

server/etc/regions.json
```diff
@@ -0,0 +1,12 @@
+[
+    {
+        "id": "f6db2e41-040e-47c7-8bba-a345b6d35ca1",
```
tahb, line 3: Is this the primary key for the probation_region? If so, this would only work on production.

In the absence of a global region ID shared between services we could either use the name "Greater Manchester" or [the delius_code ](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/223fb445d597c5723fa4c375bcfe8b58b243224f/src/main/resources/db/migration/all/20230214123554__correct_delius_code_for_greater_manchester_region.sql#L2).

tahb (commented):
In my view the API should not be responsible for providing UI control flags. Rather it should be the responsibility of the frontend to decide how and when to present data to its users.  

As things stand each of the three frontends are free to make their own decisions without being coupled to each other or changes in the API. Our current approach is to use frontend helpers (eg [1, 2]) to manage this and so I can’t foresee that the API will start returning config that this new file is standing in for.

 In our case we want to present some data differently for the Greater Manchester region. For the CAS3 user group they want to change their orientation from PDU to local authority. This change makes sense to them but may not make sense for other CAS3 user groups, the CAS1 frontend or any other consumer of the API when they interact with this region. The attempt to store and distribute this knowledge about GM centrally is well intended but I think it’s flawed as it may not always be applicable to all and it makes it harder to change. 
 
   The CAS3 frontend has confidence in its user groups, the needs of each and knows which pages need to change to accommodate them. If users start telling us that they are migrating back to PDUs and they need both columns we can easily support that with the frontend in control. 
   
I think a view helper or table presenter object which encapsulates the knowledge of what to show each region is the simplest and more conventional way forward here.

  [1] https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/684/commits/572a8ef7583d7b53b35a31579840923985153e36
   [2] https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/main/server/utils/premisesUtils.ts#L10C7-L10C15

server/etc/regions.json
andrewcharnley: The delius code doesn't come back with the user region object.

andrewcharnley: https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/main/server/utils/premisesUtils.ts#L10C7-L10C15 isn't the same, it contains no user differentiation.

The API only has one consumer and its forked derivative. It's already differing its behaviour by requiring CAS1 and CAS3 flags on calls, so I don't see where the idea it shouldn't provide CASn UI meta comes from. Besides, even if another consumer came along it would still need to know the user-group preference is to show the LA over the PDU.

If there are other id's for Greater Manchester they can be added to the regions.json. It's poor but it beats behaviour changing on a visual string.

tahb: > The delius code doesn't come back with the user region object.

We could propose that that happens. Though, Delius codes are still just strings set by Delius. 

We sadly don't have a universal ID for a region but perhaps we could set about making one.

> The API only has one consumer and its forked derivative. It's already differing its behaviour by requiring CAS1 and CAS3 flags on calls

As things stand there are three forked frontends as well as the integrations service which pulls information. There is no guarantee the architecture will remain this way though so we should be open to new consumers in future.

andrewcharnley: I like the API design, it's almost like it's been designed to stop configuration leakage on purpose. 

Eventually the inevitable will happen here, the API, which is already CAS3 specific, will end up holding some CAS3 flags. Next week a user says 'I used to be able to find a customer when I needed to (for example called Janet) but now I don't know what page they're on; I only see 10 records and have to click lots of times to find them. Please can I see 20 records instead?

tahb: On further investigation the delius codes do not have a 1:1 mapping to a probation_region, so we can't use them. Multiple delius codes _can_ refer to the same region via probation_region_mappings.

We could add a new ID to the probation_regions table however it'd just be another string with a more machine like syntax eg. "east-midlands".

```
id                                   name                   ap_area_id                           delius_code
------------------------------------ ---------------------- ------------------------------------ -----------
0544d95a-f6bb-43f8-9be7-aae66e3bf244 East Midlands          37ceb2bf-c588-4f08-ab62-8f3e7f9b0c0f N53
ca979718-b15d-4318-9944-69aaff281cad East of England        858a01cf-9fd1-482c-a983-1a74713f9227 N56
db82d408-d440-4eb5-960b-119cb33427cd Kent, Surrey & Sussex  858a01cf-9fd1-482c-a983-1a74713f9227 N57
d73ae6b5-041e-4d44-b859-b8c77567d893 London                 c482836a-bd2a-4cce-be6f-c5d548f621f8 N07
c5acff6c-d0d2-4b89-9f4d-89a15cfa3891 North East             6025b8c6-883f-4300-bc26-2b01fbd6e272 N54
a02b7727-63aa-46f2-80f1-e0b05b31903c North West             93a59a2d-1730-4154-8e02-aee6f2f01a76 N51
6b4a1308-17af-4c1a-a330-6005bec9e27b South Central          2cc6bc0f-58cf-4d82-99dd-dfc67b7f5c33 N59
43606be0-9836-441d-9bc1-5586de9ac931 South West             2cc6bc0f-58cf-4d82-99dd-dfc67b7f5c33 N58
afee0696-8df3-4d9f-9d0c-268f17772e2c Wales                  56d78348-64d4-4af2-a13c-e5c0d4d30dd0 N03
734261a0-d053-4aed-968d-ffc518cc17f8 West Midlands          37ceb2bf-c588-4f08-ab62-8f3e7f9b0c0f N52
5e44b880-df20-4751-938f-a14be5fe609d Yorkshire & The Humber 6025b8c6-883f-4300-bc26-2b01fbd6e272 N55
f6db2e41-040e-47c7-8bba-a345b6d35ca1 Greater Manchester     93a59a2d-1730-4154-8e02-aee6f2f01a76 N50
```

Given these names are stored as data and are not modifiable by users they are already acting as our universal identifiers and so can continue to be relied upon.

tahb (changes requested):
I've reviewed this and the differences of opinion and sided with the use of view helpers instead of configuration. [This choice appears to have been further challenged](https://mojdt.slack.com/archives/C03K0HB0LBE/p1701773403246079). It seems we are now at an impasse and so this work will be rewritten by someone else.

server/etc/regions.json
andrewcharnley: Here'a ticket from the board, it's PDU not LA, but you get the idea. I accept that time is pressing but someone has to maintain this service in the future and once you put this anti-pattern into play it'll be accepted and not challenged.

Screenshot 2023-12-05 at 17 19 56

I've been doing this a long time Tom, I've seen it all before.

andrewcharnley: The thing is by just adding one more column into that table all the problems disappear. You're thinking it's UI specific - it's not. A region manages properties by LA not PDU. This is a data attribute, and maybe it's required by the API itself, i.e when it generates reports? It not today, perhaps tomorrow - design in the flexibility from day 1.

PR #768, opened 2023-11-28
fix: attempt to fix e2e against current (at time of writing) dev api

andrewcharnley:
The e2e test that is failing expects a premises to exist, I rejigged the order. I think the db isn't seeded. I figure it will on the first successful e2e run, but then any dev on the api side could be wiping it.

Although it shouldn't matter, to resolve any other tests failing I sequenced the files, cypress will run them in order.


libuk:
Could you give more context on what the issue is and what this fix is attempting to do?

It also might be worth rolling back the change that caused it.

PR #772, opened 2023-11-29
Revert the pagination revert and fix all e2e pipeline issues required to make it pass

andrewcharnley:
Working against current api dev on local are the following changes.

1. I've observed a race condition which occasionally failed one test.
2. The reports test was hard coded to a specific number and would fail.
3. The bulk bookings has been reduced to the minimum required to verify pagination.

Screenshot shows e2e tests running local. I ran 7 runs consecutively with success. Hopefully the pipeline API will play ball.

Screenshot 2023-11-29 at 14 32 57


e2e/tests/stepDefinitions/place.ts
```diff
@@ -53,6 +53,15 @@ Given('I mark the assessment as ready to place', () => {
       ...assessment,
       status: 'in_review',
     })
+
+    /*
+     * Resolve a race between the links parent not showing yet.
```
tahb, line 58: Looks like [the last fix commit here is fixing three things](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/772/commits/1b0aeb8791aa2088df54d13a853585b8c5a02d87) and [as mentioned](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/764) we look to break commits like this apart. 

The first two for reportNew and making fewer bookings look reasonable. Making 4 less bookings should shave some time off and we'll have to hope it's enough 👍 The third on the race condition surprises me and it'd be good to consider further.

As we're primarily focused on pagination here this sounds unrelated, presumably it's an existing issue? Do we have a link to the output of the failure we think is a race? Could we find any evidence of it coming up before in any temperamental test failures? Should we be wary of this happening again on the other assessment state changes too? Do we think a short sleep is the best fix for this? 

I would expect the API to be creating these records synchronously and so should only be returning when it's done and successful. Is the frontend perhaps sending this request asynchronously?

tahb (commented):
Looks like we can interactive rebase two of these commits away and then break down the fix into the three things being fixed. 

As mentioned below the last race condition fix is not something I think we should rush though unless we have the answers, it would be problematic to have waits dotted about and a slower pipeline. The other two fixes however seem reasonable and are good to merge.

If the race condition was existing and unrelated to this change I think we drop that from this PR to unblock it and get it deployed. We can circle back to take our time to think about it.

e2e/tests/stepDefinitions/place.ts
andrewcharnley: The race condition only occurs on my local as far as I'm aware (I've not seen it in Circle pipeline). This suggests the additional latency between the UI and API in CI is allowing the API to fulfil its backend op. If the UI was firing async I'd definitely expect it to come up in CI.

andrewcharnley: I'll put this one on-ice until I hear which approach to take.

adamhughes86: Yeah lets keep the race condition on the back burner for now. I would have thought it would have been raised earlier if it was a widespread issue.

If you see it happening again then we'll take a look at what we can do to resolve it, preferable without relying on waits. It would be good to fully understand it.

As a thought - You could possible try and update the cypress default wait value as long as that wasn't committed up. Doesn't fix the actual issue but is a work around

PR #777, opened 2023-11-29
feat: Add booking pagination (but this time e2e relying on API seed records).

andrewcharnley:
Requirements and story for this ticket can be found at

https://trello.com/c/vhrk6bcy/1684-booking-search-timing-out-when-searching-arrived-and-departed-bookings

No changes on the originally merged PR which was reversed to get another in before it (the e2e test was too slow). The only change here is to switch to API seeded records.


richpjames (approved):
I'd echo Dan and Tom's comments that the original PR was far too large for a single commit but given that it was merged I think it's reasonable that this PR is too

PR #791, opened 2023-12-12
Revert "Revert the revert of "Add booking pagination"" so it's coming out

tahb:
Reverts ministryofjustice/hmpps-temporary-accommodation-ui#789

[It seems like this feature when on prod with more data has come up against an existing query. We're trying to fix the query but we may need to revert this (again) until that can be addressed.](https://mojdt.slack.com/archives/C03K0HB0LBE/p1702383217041969).


pezholio (approved):
Reverts make me sad, but LGTM 👍

PR #792, opened 2023-12-12
Revert "Revert "Revert the revert of "Add booking pagination"" so it's going in

tahb:
Reverts ministryofjustice/hmpps-temporary-accommodation-ui#791

[We had a production incident and this looked like a related change](https://mojdt.slack.com/archives/C03K0HB0LBE/p1702383217041969) so we reverted it. 

It turns out the problem persisted so is unlikely down to this feature so when we're ready it can go back. [We should however perhaps wait for the default page size to be increased from 10 to 50/100 in response to user feedback.](https://trello.com/c/qCepzjiu)


froddd:
Superseded by #896

PR #875, opened 2024-02-26
FIX: Ensure test fixture town isn't regenerated when parsing address

froddd:
# Context

When running E2E tests which select an active premises to view and act on, the step `'I view an existing active premises'` would set the test suite's active premises based on the data available on the page, but ignoring the town. This meant the town name was regenerated for every test attempt, and therefore failing.

# Changes in this PR

This updates the `parsePremises` method to take into account the town as shown on the page, therefore making this data consistent.

The `town` property is set based on the penultimate line of the address block on the page, with the postcode being the last. The `addressLine2` property is optional, and while not seemingly used in any of the test data, it is safer to ensure the town is parsed as the penultimate line, as it is always set, as is the postcode.

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/cypress_shared/pages/temporary-accommodation/manage/premisesShow.ts
+++ b/cypress_shared/pages/temporary-accommodation/manage/premisesShow.ts
@@ -181,6 +181,7 @@ export default class PremisesShowPage extends Page {
       id,
       name,
       addressLine1: addressLines[0],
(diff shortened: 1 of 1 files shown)


sudhakar-plr (approved):
Looks ok to me

PR #885, opened 2024-03-04
CAS-174 split referral requests

froddd:
# Context

https://dsdmoj.atlassian.net/browse/CAS-174

The Referrals page (`/review-and-assess`) currently calls the API to fetch _all_ referrals of any status, including referrals that should be shown as archived (statuses `closed` and `rejected`). These referrals are then split into 'Unallocated', 'In review', and 'Ready to place' in the front-end; the same call is used to show the 'Archived referrals' page.

The current page performance is poor, with API queries up to around 6 minutes (!). We expect performance to be drastically improved when the query targets only the relevant status(es) for each of these pages.

This PR splits the 'Unallocated', 'In review' and 'Ready to place' views into 3 separate pages, each requesting a filtered set of referrals from the API. The 'Archived referrals' page now also targets only the specific statuses that are needed to be displayed.

# Changes in this PR

As a result of splitting the data into 3 separate pages, the Tabs GOVUK component is no longer appropriate, and has been replaced with the sub-navigation one. This results in more of the screen real-estate being used, and is consistent with the Bookings search screens.

The title is now also page-specific.

In order to prevent existing bookmarks or links from breaking, a permanent redirect takes users from the previous screen, `/review-and-assess`, to the new 'Unallocated referrals' URL, `/review-and-assess/unallocated`.

### Before

Screenshot 2024-03-05 at 15 35 48

### After

Screenshot 2024-03-05 at 15 34 00

# Release checklist

## Pre-merge checklist

- [ ] Are any changes required to dependent services for this change to work? Yes: https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/1568
- [ ] Have they been released to production already? No

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/server/@types/ui/index.d.ts
+++ b/server/@types/ui/index.d.ts
@@ -324,3 +324,5 @@ export type ApplicationSummaryData = {
     | 'CAS2 (formerly Bail Accommodation Support Services)'
     | null
 }
+
+export type AssessmentSearchApiStatus = 'unallocated' | 'in_review' | 'ready_to_place'
--- a/server/@types/ui/index.d.ts
+++ b/server/@types/ui/index.d.ts
@@ -8,6 +8,7 @@ import {
   ArrayOfOASysRiskToSelfQuestions,
   ArrayOfOASysSupportingInformationQuestions,
   TemporaryAccommodationAssessment as Assessment,
+  AssessmentStatus,
   Booking,
   Document,
   LocalAuthorityArea,
@@ -325,4 +326,4 @@ export type ApplicationSummaryData = {
     | null
 }
 
-export type AssessmentSearchApiStatus = 'unallocated' | 'in_review' | 'ready_to_place'
+export type AssessmentSearchApiStatus = Extract<AssessmentStatus, 'unallocated' | 'in_review' | 'ready_to_place'>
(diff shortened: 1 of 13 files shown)


server/@types/ui/index.d.ts
```diff
@@ -324,3 +324,5 @@ export type ApplicationSummaryData = {
     | 'CAS2 (formerly Bail Accommodation Support Services)'
     | null
 }
+
+export type AssessmentSearchApiStatus = 'unallocated' | 'in_review' | 'ready_to_place'
```
richpjames: Could you use this type and exclude the statuses that aren't needed? https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/src/main/resources/static/_shared.yml#L2400 
That way the statuses will always be in sync with the ones in the API

server/data/assessmentClient.test.ts
```diff
@@ -33,15 +34,23 @@ describe('AssessmentClient', () => {
...
-      fakeApprovedPremisesApi
-        .get(paths.assessments.index({}))
-        .matchHeader('authorization', `Bearer ${callConfig.token}`)
-        .reply(200, assessmentSummaries)
-
-      const output = await assessmentClient.all()
-      expect(output).toEqual(assessmentSummaries)
-    })
+    it.each([
+      ['unallocated', 'unallocated' as AssessmentStatus],
```
richpjames: possibly not, but would casting these `as const` allow TS to infer them as `AssessmentStatus`es?

richpjames (dismissed):
LGTM, have left a couple of minor comments

server/@types/ui/index.d.ts
froddd: Yup. I guess using `Extract` here may make more sense, as with `Exclude` any extra literals added to the API type would leak into this.

richpjames (approved):

PR #928, opened 2024-03-28
CAS-204 domain events content

froddd:
# Context

https://dsdmoj.atlassian.net/browse/CAS-204

Certain domain events are now being fired which remove the need for users to manually update nDelius after certain actions. This change updates or removes messaging asking users to update nDelius.

# Changes in this PR

- When marking a booking as 'Arrived' or 'Departed', a message warning the user to update nDelius above the submit button is now removed.
- When a booking is marked as 'Arrived' or 'Departed', success message now says 'You no longer need to update nDelius with this change.'

### Before

Screenshot 2024-03-28 at 16 22 08

### After

Screenshot 2024-03-28 at 16 18 17

# Release checklist

[Release process documentation](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/edit-v2/4247847062?draftShareId=a1c360ab-bd31-4db1-aae3-7cc002761de9).

## Pre-merge checklist

- [ ] Are any changes required to dependent services for this change to work?
  (eg. CAS API)
- [ ] Have they been released to production already?

## Post-merge checklist

- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to test
- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to preprod
- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to prod

PR #929, opened 2024-04-02
CAS-204: Avoid hyphens in env var

froddd:
It seems env vars with hyphens are simply ignored. Given the fallback for the offending one (#928), this results in the actual value set in Helm chart not being applied.


sudhakar-plr (approved):
LGTM

PR #934, opened 2024-04-08
CAS-60 show days from now

froddd:
# Context

https://dsdmoj.atlassian.net/browse/CAS-60

# Changes in this PR

This adds extra descriptive text next to the bedspace end date, showing ` (X days from now)`. 

The new `DateFormats.isoDateToDaysFromNow` method is able to render days in the past, too, as filtering fora future dates only would have been quite prescriptive and opaque. The logic to show only future date is therefore kept separate.

## Screenshots of UI changes

Screenshot 2024-04-08 at 11 34 48

### DIFF
--- a/server/utils/dateUtils.ts
+++ b/server/utils/dateUtils.ts
@@ -1,6 +1,6 @@
 /* eslint-disable */
 import type { ObjectWithDateParts } from '@approved-premises/ui'
-import { isFuture, isPast } from 'date-fns'
+import { differenceInDays, isFuture, isPast } from 'date-fns'
 
 import format from 'date-fns/format'
 import formatISO from 'date-fns/formatISO'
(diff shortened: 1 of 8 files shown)


server/utils/dateUtils.ts
```diff
@@ -119,6 +119,20 @@ export class DateFormats {
       [`${key}-year`]: `${date.getFullYear()}`,
     } as ObjectWithDateParts<K>
   }
+
+  static isoDateToDaysFromNow(dateString: string) {
+    const difference = differenceInDays(new Date(dateString).setHours(0, 0, 0, 0), new Date().setHours(0, 0, 0, 0))
```
froddd, line 124: `date-fns`'s `differenceInDays` returns the number of complete 24hr cycles between dates: this means it finds 0 days between `2024-04-08T12:00:00` and `2024-04-09T00:00:00`. To avoid this pitfall, as we are only concerned with whole dates, we set both compared dates to `T00:00:00.000`.

richpjames (approved):

PR #954, opened 2024-04-19
Add bedspace reference to booking screens

marcocolluramoj:
# Context

https://dsdmoj.atlassian.net/browse/CAS-171

# Changes in this PR

Adds the bedspace reference with the property address on booking screens.

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/server/views/temporary-accommodation/components/location-header/macro.njk
+++ b/server/views/temporary-accommodation/components/location-header/macro.njk
@@ -10,10 +10,10 @@
     {% if not hideAddress %}
       <h2>Property address</h2>
       <p>
-            {{ room.name }}<br />
-            {{ premises.addressLine1 }}<br/>
-            {% if premises.town.length %}{{ premises.town }}{% endif %}<br />
-            {{ premises.postcode }}
+            {% if room.name.length %}{{ room.name }}<br/>{% endif %}
+            {{ premises.addressLine1 }}
+            {% if premises.town.length %}<br/>{{ premises.town }}{% endif %}
+            <br />{{ premises.postcode }}
       </p>
     {% endif %}
   {% endif %}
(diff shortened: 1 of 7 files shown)


server/views/temporary-accommodation/components/location-header/macro.njk
```diff
@@ -4,13 +4,17 @@
...
     {% if not hideAddress %}
       <h2>Property address</h2>
-      <p>{{ premises.addressLine1 }}
-            {% if premises.town.length %}<br/>{{ premises.town }}{% endif %}
-            <br />{{ premises.postcode }}</p>
+      <p>
+            {{ room.name }}<br />
+            {{ premises.addressLine1 }}<br/>
+            {% if premises.town.length %}{{ premises.town }}{% endif %}<br />
+            {{ premises.postcode }}
```
froddd: I think we're losing the original logic which would only add a line break where needed if the town was shown -- the same approach should be used for the `{{ room.name }}` line, I guess. Here's what it looks like with the extra blank line rendered at the start:

Screenshot 2024-04-22 at 12 38 09



To ensure we're not adding blank lines when there is no data to show:

```suggestion
            {% if room.name.length %}{{ room.name }}<br />{% endif %}
            {{ premises.addressLine1 }}<br/>
            {% if premises.town.length %}{{ premises.town }}<br />{% endif %}
            {{ premises.postcode }}
```

froddd (changes requested):
Small issue with extra blank line being rendered when there's no room ID to show.

server/views/temporary-accommodation/components/location-header/macro.njk
marcocolluramoj: Thanks @froddd I've added the conditional line breaks back in and done the same for room.name

froddd: Seems to have sorted it 👍

froddd (approved):
LGTM!

PR #961, opened 2024-04-25
Attempt to speed up E2Es

pezholio:
I've noticed a couple of bottlenecks when running the E2Es, particularly in the API pipeline. This should hopefully speed things up a tad.


froddd (approved):
Nice one, looks good!

PR #963, opened 2024-04-26
Revert "Use cy.session to share login sessions"

pezholio:
This is causing issues. I think there might be some more config needed to get it working with Cucumber


davidatkinsuk (approved):
LGTM

PR #971, opened 2024-05-13
BUGFIX: Ensure years are 4 digits

froddd:
# Changes in this PR

Ensure date validation considers years with 3 digits invalid, to prevent users accidentally entering incomplete years like `202` instead of `2024`

### DIFF
--- a/server/utils/dateUtils.ts
+++ b/server/utils/dateUtils.ts
@@ -41,7 +41,13 @@ export class DateFormats {
    * @throws {InvalidDateStringError} If the string is not a valid ISO8601 datetime string
    */
   static isoToDateObj(date: string) {
-    const parsedDate = new Date(date)
+    let parsedDate: Date
+
+    try {
+      parsedDate = new Date(date)
+    } catch (error) {
+      throw new InvalidDateStringError(`Invalid Date: ${date}`)
+    }
 
     if (Number.isNaN(parsedDate.getTime())) {
       throw new InvalidDateStringError(`Invalid Date: ${date}`)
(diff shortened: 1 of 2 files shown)


richpjames (approved):

server/utils/dateUtils.ts
```diff
@@ -153,6 +159,10 @@ export const dateAndTimeInputsAreValidDates = <K extends string | number>(
   dateInputObj: Partial<ObjectWithDateParts<K>>,
   key: K,
 ): boolean => {
+  const inputYear = dateInputObj?.[`${key}-year`] as string
+
+  if (inputYear && inputYear.length !== 4) return false
```
richpjames, line 164: Before this were 2 digit years valid (EG '24)? I'm not sure off the top of my head whether that would parse but if so we should probably retain that behaviour

froddd: Approved Premises UI uses `parseISO` from `date-fns` to parse, and that fails on years with 2 digits (`parseISO('99-01-01') -> Invalid Date`) so I'm not sure why that check was there anyway in Approved Premises UI, but in Temporary Accommodation UI we're now using `new Date()` instead, which considers dates with 2-digit years valid, but renders those completely wrong so best avoid them:

```
new Date('01-01-01') // 2001-01-01T00:00:00.000Z, ok
new Date('99-01-01') // 1999-01-01T00:00:00.000Z, ok
new Date('03-05-01') // 2001-03-05T00:00:00.000Z, lolwut?!
```

PR #989, opened 2024-06-19
CHORE/Cleanup unnecessary custom type

froddd:
Rely on existing API-generated type for assessment rejection body instead of custom type, as these offer sufficient overlap.


danhumphreys-moj (approved):
`AssessmentRejection` is in the API `_shared.yml` [here](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/116f3066ac4bc37720eb0586296b06ba47201b47/src/main/resources/static/_shared.yml#L2737).

PR #991, opened 2024-06-19
CHORE/update dependencies 2024 06 19

froddd:
Update all non-major dependencies.

Integrations provided by `@sentry/tracing` are now included in the main `@sentry/node` package, so this dependency has been removed and the integration updated.


vubalasu:
@froddd better to squash all the commits

vubalasu (approved):

PR #995, opened 2024-07-01
Use uppercase for local e2e test usernames

froddd:
This change is needed after a [recent change to the API](https://github.com/ministryofjustice/hmpps-approved-premises-tools/pull/69) in how it handles nDelius users for CAS1 E2E tests.


froddd:
Fixed in API stubs: https://github.com/ministryofjustice/hmpps-approved-premises-tools/pull/75

PR #1006, opened 2024-07-11
Simplify e2e alert template

froddd:
An attempt to simplify the template used for Slack notifications for CircleCI e2e test failures. This new templates follows [the one from the Ministry of Justice CircleCI orb](https://github.com/ministryofjustice/hmpps-circleci-orb/blob/main/src/files/slack_failed.tpl.json) more closely, but replaces one of the duplicated job statuses with the Author of the ref.

This also fully removes the `check_outdated` job, which is no longer referenced anywhere in the config.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,28 +18,6 @@ parameters:
     default: 18.16-browsers
 
 jobs:
-  check_outdated:
-    executor:
-      name: hmpps/node
-      tag: << pipeline.parameters.node-version >>
-    steps:
-      - checkout
-      - restore_cache:
-          key: dependency-cache-{{ checksum "package-lock.json" }}
-      - run:
-          name: install-npm
-          command: 'npm ci --no-audit'
-      - run:
-          name: Check version
-          command: 'npm --version'
-      - run:
-          name: Run check
-          command: 'npm outdated typescript govuk-frontend'
-      - slack/notify:
-          event: fail
-          channel: << pipeline.parameters.alerts-slack-channel >>
-          template: basic_fail_1
-
   e2e_environment_test:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -72,7 +72,35 @@ jobs:
       - slack/notify:
           event: fail
           channel: << pipeline.parameters.alerts-slack-channel >>
-          template: basic_fail_1
+          custom: |
```
davidatkinsuk, line 53: given we're using the hmpps circle ci orb, is there no way to use their template definition directly (slack_failed.tpl.json)?

Looking at the following example i wonder what happens if we remove the template definition altogether - will it fall back to version in hmpps circle ci orb?

or maybe we can refer to the hmpps templates in the template param e.g. slack_failed ? 

https://github.com/ministryofjustice/hmpps-circleci-orb/blob/main/src/examples/slack_notification.yml

davidatkinsuk (approved):
👍

.circleci/config.yml
froddd: Every call to the Slack Notify task in the Orb [defines the template explicitly](https://github.com/ministryofjustice/hmpps-circleci-orb/blob/main/src/jobs/npm_outdated.yml#L32) using the `<< include(template.json) >>` syntax, but this syntax is only available to Orb definitions ([docs](https://circleci.com/docs/orb-concepts/#file-include-syntax)).

I don’t think the template name is exposed beyond the orb either. The Orb’s deploy success task actually defines its own template in place [here](https://github.com/ministryofjustice/hmpps-circleci-orb/blob/main/src/jobs/deploy_env.yml#L130), which makes me think it has to be defined explicitly.

davidatkinsuk: Cool, thanks for looking into it

PR #1007, opened 2024-07-12
Move subheading to be included in the h1

aliuk2012:
# Context

- https://dsdmoj.atlassian.net/browse/CAS-471

# Changes in this PR

Subheading in the context they are being used would be very important for screenreaders to inform users what page they are viewing.

Example, If this change was not done then screenreaders would only read out "Joe Bloggs Unallocated" instead of  "Referral summary Joe Bloggs "

ref: https://design-system.service.gov.uk/styles/headings/#headings-with-captions
If the caption should be considered part of the page heading, you can also nest the caption within the <h1>.

There should be no visional difference from what users previously saw.

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/server/views/temporary-accommodation/assessments/summary.njk
+++ b/server/views/temporary-accommodation/assessments/summary.njk
@@ -33,9 +33,11 @@
         <div class="moj-identity-bar__container">
             <div class="moj-identity-bar__details">
         <span class="moj-identity-bar__title">
-          <span class="govuk-caption-xl">Referral summary</span>
-          <h1 class="govuk-heading-xl">{{ name }}
-              {{ assessmentStatusTag(assessment.status) | safe }}</h1>
+            <h1 class="govuk-heading-xl">
+                <span class="govuk-caption-xl">Referral summary</span>
+                {{ name }}
+                {{ assessmentStatusTag(assessment.status) | safe }}
+            </h1>
         </span>
             </div>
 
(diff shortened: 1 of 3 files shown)


server/views/temporary-accommodation/assessments/summary.njk
```diff
@@ -33,9 +33,11 @@
         <div class="moj-identity-bar__container">
             <div class="moj-identity-bar__details">
         <span class="moj-identity-bar__title">
-          <span class="govuk-caption-xl">Referral summary</span>
-          <h1 class="govuk-heading-xl">{{ name }}
-              {{ assessmentStatusTag(assessment.status) | safe }}</h1>
+            <h1 class="govuk-heading-l">
+                <span class="govuk-caption-l">Referral summary</span>
```
froddd: I don't think the change of size modifier from `-xl` to `-l` is desirable here? This does change the rendered size.

froddd (changes requested):

server/views/temporary-accommodation/assessments/summary.njk
aliuk2012: Fixed now.

froddd:
Just came across this extra caption which should also be moved into the following heading: https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/801f1fa30e9f48ebf503ea4923dacf08fa3afaf0/server/views/applications/full.njk#L28

froddd (approved):
👍

PR #1010, opened 2024-07-15
CAS-477 Relocate all components view to server views components

aliuk2012:
# Context

# Changes in this PR

Relocate all components from `server/views/temporary-accommodation/components` to `server/views/components`.
This is to help us have one location for all view components and better control over what we class as a view component.

Once this is merged, all developers should manually delete `server/views/temporary-accommodation/components` as it will still be present but will not contain any components


froddd:
Given there are 2 versions of a few of the form field macros, would it be worth reducing the duplication and removing those that already exist in another form?

froddd (approved):
👍

PR #1012, opened 2024-07-16
CAS-478 remove duplicate form field components

aliuk2012:
# Context

# Changes in this PR

Replaces the following ta components with forms-field components which seems to be used across all the CAS 

| ta            | formFields     |
|----------------------|---------------|
| ta-textarea   | form-page-textarea  |
| ta-input        | form-page-input      | 
| ta-date-input | form-page-date-input |

The only form field components that are still duplicated are the radio buttons and checkboxes. This is because both of these components work very differently. In ta versions the checkboxes/radio buttons are calculated at controller level where as the form-field components calculate in the component itself


server/views/components/formFields/form-page-date-input/macro.njk
```diff
@@ -5,7 +5,7 @@
       govukDateInput(
         mergeObjects(
           params,
-          { id: params.fieldName, namePrefix: params.fieldName, errorMessage: context.errors[params.fieldName], value: context[params.fieldName] }
+          { id: params.fieldName, name: params.fieldName, errorMessage: context.errors[params.fieldName], value: context[params.fieldName] }
```
froddd: What's the reason behind this change? It seems the `govukDateInput` does expect `namePrefix` rather than `name`

froddd: This seems to break functionality, as the resulting inputs simply have `name` attributes set to `day`, `month`, and `year`, which means their values aren't submitted to the expected form data

aliuk2012: 🤦  I could have sworn one of those components mentioned namePrefix which was incorrect. might of been another component. I've reverted these changes now

froddd (approved):
LGTM!

PR #1013, opened 2024-07-18
Replace `exclude-from-print` css class name with govuk util class govuk-\!-display-none-print

aliuk2012:
# Context
Should be no visible changes to the user and it means we are not having to maintain our own custom classes when they are already being included by govuk frontend by default

- https://dsdmoj.atlassian.net/browse/CAS-476

### DIFF
--- a/server/utils/checkYourAnswersUtils/index.ts
+++ b/server/utils/checkYourAnswersUtils/index.ts
@@ -47,7 +47,7 @@ const summaryListItemForResponse = (
         href: paths.applications.pages.show({ task: task.id, page: pageName, id: application.id }),
         text: 'Change',
         visuallyHiddenText: key,
-        classes: 'exclude-from-print',
+        classes: 'govuk-!-display-none-print',
       },
     ],
   }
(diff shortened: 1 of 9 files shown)


server/utils/checkYourAnswersUtils/index.ts
```diff
@@ -47,7 +47,7 @@ const summaryListItemForResponse = (
         href: paths.applications.pages.show({ task: task.id, page: pageName, id: application.id }),
         text: 'Change',
         visuallyHiddenText: key,
-        classes: 'exclude-from-print',
+        classes: 'govuk-\\!-display-none-print',
```
froddd: Those backslashes (escaping?) shouldn't be needed; they result in the class name being rendered as `govuk-\!-display-none-print`, which means the links aren't actually hidden when trying to print!

aliuk2012: Should be fixed now

froddd (approved):
👍

PR #1015, opened 2024-07-19
Fix pages with single date inputs

aliuk2012:
The current implementation causes a lot of accessibility problems for screen readers as the screenreader does not have context as to what is being asked from the user. All it will read out is "Day", "Month", or "Year"  depending on which input has focus. We should be using IsPageHeading option so that a fieldset is placed around the date field but it also injects an h1 tag inside the fieldsets legend. A screen will then announce the legend to the user so the will have the question and hint

There was an accessibility issue reported as part of an assessment done late 2023 which read:

>Date input fields are not grouped with questions, so screen reader users may find it challenging to associate the date input field with the relevant question (test issue 1.3). This fails WCAG 2.2 success criterion 1.3.1 (A). We plan to fix this by 30/03 2024.

This commit should mean this issue is now fixed across the service.

# Context

# Changes in this PR

### Before
Screenshot 2024-07-18 at 16 19 06

### After
Screenshot 2024-07-18 at 16 19 14

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/server/views/applications/pages/accommodation-need/eligibility/accommodation-required-from-date.njk
+++ b/server/views/applications/pages/accommodation-need/eligibility/accommodation-required-from-date.njk
@@ -2,13 +2,20 @@
 
 {% block questions %}
 
-  <h1 class="govuk-heading-l">
-    <span class="govuk-caption-l">{{ task.title }}</span>
-    {{ page.title }}
-  </h1>
+{% set pageTitleHTML %}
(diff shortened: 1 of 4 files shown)


Rosie-Brigham (approved):
👍

PR #1016, opened 2024-07-23
Update single question pages to use h1 as legend

aliuk2012:
# Context

https://dsdmoj.atlassian.net/browse/CAS-500

This is needed for accessibility reasons to ensure that radio button/checkbox options have enough context for the screen reader users to understand what is needed by making the heading part of a fieldset legend.


Rosie-Brigham (approved):
👍

PR #1020, opened 2024-07-25
Explicitly override PDU for dev e2e user

froddd:
The user the E2E tests are based on is in a different region to the local user used for testing. As the probation region is now tested based on the user's region, the PDU data used for local E2E tests cannot be used for the dev E2E tests.

This PR overrides the default application data when running on dev, in order for the tests to pass.


davidatkinsuk (approved):
LGTM 👍

PR #1021, opened 2024-07-29
CAS-270 Rename search crn param to crnOrName

aliuk2012:
# Context
https://dsdmoj.atlassian.net/browse/CAS-270

# Changes in this PR
We are changing how search works and in do so the search param will no longer only search for crn but name as well. This commit is simple renaming the existing param so its clear that its not only for crn

### Before
image

 
image

### After
image

image

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/integration_tests/tests/temporary-accommodation/assess/assess.cy.ts
+++ b/integration_tests/tests/temporary-accommodation/assess/assess.cy.ts
@@ -96,31 +96,31 @@ context('Apply', () => {
         const pageTitle = sentence(`${uiStatus} referrals`)
 
         describe(`when viewing ${uiStatus} referrals`, () => {
-          it('shows the result of a CRN search and clears the search', () => {
+          it('shows the result of a CRN or Name search and clears the search', () => {
             // Given there are assessments in the database
             const assessments = assessmentSummaryFactory.buildList(6, { status: factoryStatus })
             const searchedForReferral = assessments[1]
-            const searchCRN = searchedForReferral.person.crn
+            const searchCrnOrName = searchedForReferral.person.crn
             cy.task('stubAssessments', { data: assessments, pagination })
 
             // When I visit the Archived referrals page
             const page = ListPage.visit(status)
 
-            // Then the search by CRN form is empty
-            page.checkCRNSearchValue('', uiStatus)
+            // Then the search by CRN or Name form is empty
+            page.checkCrnOrNameSearchValue('', uiStatus)
 
             // And I see all the results
             page.checkResults(assessments)
 
-            // When I submit a search by CRN
+            // When I submit a search by CRN or Name
             cy.task('stubAssessments', { data: [searchedForReferral] })
-            page.searchByCRN(searchCRN, uiStatus)
+            page.searchByCrnOrName(searchCrnOrName, uiStatus)
             Page.verifyOnPage(ListPage, pageTitle)
 
-            // Then the search by CRN form is populated
-            page.checkCRNSearchValue(searchCRN, uiStatus)
+            // Then the search by CRN or Name form is populated
+            page.checkCrnOrNameSearchValue(searchCrnOrName, uiStatus)
 
-            // Then I see the search result for that CRN
+            // Then I see the search result for that CRN or Name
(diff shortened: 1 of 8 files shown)


server/testutils/factories/assessmentSearchParameters.ts
```diff
@@ -3,5 +3,5 @@ import { AssessmentSearchParameters } from '@approved-premises/ui'
 import { faker } from '@faker-js/faker/locale/en_GB'
 
 export default Factory.define<AssessmentSearchParameters>(() => ({
-  crn: `C${faker.number.int({ min: 100000, max: 999999 })}`,
+  query: faker.lorem.lines(),
```
froddd: This would generate several lines of text, which seems counterintuitive here -- wouldn't this best generate either a CRN or a first name, for instance?
```suggestion
  query: faker.helpers.arrayElement([
    `C${faker.number.int({ min: 100000, max: 999999 })}`,
    faker.person.firstName(),
  ]),
```

server/controllers/temporary-accommodation/manage/assessmentsController.ts
```diff
@@ -75,13 +75,7 @@ export default class AssessmentsController {
...
 
       try {
-        if (params.crn !== undefined && !params.crn.trim().length) {
-          const error = new Error()
-          insertGenericError(error, 'crn', 'empty')
-          throw error
-        }
-
-        const response = errors.crn
+        const response = errors.query
```
froddd: If no errors are thrown then this conditional isn't needed

server/controllers/temporary-accommodation/manage/assessmentsController.test.ts
```diff
@@ -254,10 +254,10 @@ describe('AssessmentsController', () => {
           })
         })
 
-        describe('when there is a CRN search error', () => {
+        describe('when there is a search error', () => {
```
froddd: I believe we're not showing errors, instead clearing the search when there's no search term? This test name should reflect that.

server/controllers/temporary-accommodation/manage/assessmentsController.test.ts
```diff
@@ -229,22 +229,22 @@ describe('AssessmentsController', () => {
...
               pagination: {},
-              crn: searchParameters.crn,
+              query: searchParameters.query,
               errors: {},
             })
           })
         })
 
-        describe('when a blank CRN search is submitted', () => {
+        describe('when a blank search is submitted', () => {
```
froddd: This should clearly show a blank search should clear the search (sorry, I think this links to the next line more than this one, but it wasn't linkable to this comment as it was unchanged!)

integration_tests/tests/temporary-accommodation/assess/assess.cy.ts
```diff
@@ -100,7 +100,7 @@ context('Apply', () => {
             // Given there are assessments in the database
             const assessments = assessmentSummaryFactory.buildList(6, { status: factoryStatus })
             const searchedForReferral = assessments[1]
-            const searchCRN = searchedForReferral.person.crn
+            const searchQuery = searchedForReferral.person.query
```
froddd: Will definitely need for the API to be updated and offer new types for this to be merged, but this will have to be done with a deprecation step -- i.e. API should accept both `crn` and `query` parameters, for this to be merged and deployed, then the `crn` parameter should be retired from the API, to ensure no loss of functionality without having to synchronise deploys (which isn't very doable, unfortunately!)

froddd (commented):
Left a few comments, we'll have to wait for API updates anyway on this!

server/controllers/temporary-accommodation/manage/assessmentsController.test.ts
froddd: Or be deleted altogether!

aliuk2012: I don't think this is relevant anymore

froddd (approved):
👍

PR #1022, opened 2024-07-29
CAS-366: Change ServiceNow link to open in new tab

froddd:
This ensures the user flow through an application isn't disturbed by the action.


aliuk2012 (approved):
Nice little refactor

PR #1033, opened 2024-08-07
Revert "Update dependency express-prom-bundle to v7"

aliuk2012:
Reverts ministryofjustice/hmpps-temporary-accommodation-ui#808


froddd (approved):
👍

PR #1034, opened 2024-08-07
Revert "Update dependency @types/supertest to v6" & "Update dependency @types/superagent to v8"

aliuk2012:
Reverts ministryofjustice/hmpps-temporary-accommodation-ui#802 and https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/801


froddd (approved):
👍

PR #1038, opened 2024-08-12
Remove assessment search empty query param validation

aliuk2012:
# Context

This was removed because it introduced a new "validation" pattern in the service and didn't match what GOV.UK design system recommends. On the off chance the user removes a previous search terms and submits the form they would get this copy telling them that they cant search without text when it could just be seen as "clearing" the search filter

### DIFF
--- a/server/controllers/temporary-accommodation/manage/bookingSearchController.test.ts
+++ b/server/controllers/temporary-accommodation/manage/bookingSearchController.test.ts
@@ -9,7 +9,7 @@ import { BookingSearchService } from '../../../services'
 import { bookingSearchParametersFactory } from '../../../testutils/factories'
 import extractCallConfig from '../../../utils/restUtils'
 import { convertApiStatusToUiStatus, createSubNavArr, createTableHeadings } from '../../../utils/bookingSearchUtils'
-import { catchValidationErrorOrPropogate, fetchErrorsAndUserInput, insertGenericError } from '../../../utils/validation'
+import { fetchErrorsAndUserInput } from '../../../utils/validation'
 
 jest.mock('../../../utils/restUtils')
 jest.mock('../../../utils/bookingSearchUtils')
@@ -67,7 +67,6 @@ describe('BookingSearchController', () => {
         uiStatus: 'provisional',
         tableHeadings: [],
         subNavArr: [],
-        errors: {},
         response: paginatedResponse,
         pagination: {},
       })
@@ -87,7 +86,6 @@ describe('BookingSearchController', () => {
         uiStatus: 'active',
         tableHeadings: [],
         subNavArr: [],
-        errors: {},
         response: paginatedResponse,
         pagination: {},
       })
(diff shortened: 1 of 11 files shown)


server/controllers/temporary-accommodation/manage/bookingSearchController.test.ts
```diff
@@ -162,38 +158,11 @@ describe('BookingSearchController', () => {
...
         },
       )
     })
-
-    describe('when an blank CRN search is submitted', () => {
-      it.each(['provisional', 'confirmed', 'active', 'departed'])(
-        'renders an error for %s bookings',
-        async uiStatus => {
-          const status = (uiStatus === 'active' ? 'arrived' : uiStatus) as BookingSearchApiStatus
-          const searchParameters = bookingSearchParametersFactory.build({ crn: '   ' })
```
froddd, line 178: The previous implementation trimmed the CRN to assert it had a minimum length -- this makes me wonder if we should actually do this for any input for this?

server/views/temporary-accommodation/assessments/index.njk
```diff
@@ -14,7 +14,7 @@
 
 {# TODO Move this HTML to the searchByCrnOrNameResult once HTML can be standardised between referrals and bookings #}
 {% set resultsHtml %}
-    {% if (not context.errors.prn) and ((tableRows|length > 0) or (not crn)) %}
+    {% if tableRows|length > 0 %}
```
froddd: I believe we still need `or not crn` here, as otherwise we would show 'There are no results for ‘’ in ...' if there is nothing to show (e.g. no records for that view, at all). Possibly not the best UX in this case, but I think an empty table is less confusing than a message about an empty search the user haven't even submitted!

server/views/temporary-accommodation/booking-search/results.njk
```diff
@@ -11,7 +11,7 @@
 
 {# TODO Move this HTML to the searchByCrnOrNameResult once HTML can be standardised between referrals and bookings #}
 {% set resultsHtml %}
-    {% if (not context.errors.prn) and ((response.data|length > 0) or (not crn)) %}
+    {% if response.data|length > 0 %}
```
froddd: Same as above, I think `or not crn` should still be here

froddd (changes requested):
Couple of comments for the case when there are genuinely no records but no search has been submitted.

server/views/temporary-accommodation/assessments/index.njk
aliuk2012: I'll add it back for the moment while we improve the UX.

server/views/temporary-accommodation/booking-search/results.njk
aliuk2012: I'll add it back for the moment while we improve the UX.

server/controllers/temporary-accommodation/manage/bookingSearchController.test.ts
aliuk2012: I agree. I'm going to trim the users input as part of the rename the search param and completely ignore the param if its empty and treat it as if the user clicked on "Clear"

froddd (approved):
👍

PR #1057, opened 2024-08-14
CAS-489: alternative PDU - content fixes

froddd:
# Context

https://dsdmoj.atlassian.net/browse/CAS-489

# Changes in this PR

A small content change to fix case for initialism explanation for PDU (Probation Delivery Unit), with explanation title cased as per content guidelines.


aliuk2012 (approved):
I'll make sure the changes I'm making in CAS-373 matches this too

PR #1060, opened 2024-08-15
CAS-549: update typescript

froddd:
# Context

https://dsdmoj.atlassian.net/browse/CAS-549

# Changes in this PR

Updated typescript package to version `5.5.4`, and fixed all related typechecking issues.

Removed unused types.

### DIFF
--- a/cypress_shared/pages/apply/applyPage.ts
+++ b/cypress_shared/pages/apply/applyPage.ts
@@ -17,8 +17,7 @@ export default class ApplyPage extends Page {
   ) {
     super(title)
 
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const Class = Apply.pages[taskName][pageName] as any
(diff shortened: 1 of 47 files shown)


cypress_shared/pages/apply/applyPage.ts
```diff
@@ -17,8 +17,7 @@ export default class ApplyPage extends Page {
   ) {
     super(title)
 
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
aliuk2012, line 20: 😍 Are there any other instances were we disable this linting rule?

aliuk2012 (approved):
Some nice tidying up going on here but also a few other things that could be better but I understand it was a bit of "whack-a-mole" at times but I think you have tried to make the changes as simple as possible for the time being. 

Lets go with this for the time being and we can always improve further as we go along.

Only one question about linting rule if you could look at.

PR #1066, opened 2024-08-28
run e2e tests in parallel on Circle CI

aliuk2012:
# Context
Currently it can take over 25mins for e2e tests to pass on circle ci. this is an attempt to speed it up by running the cypress tests in parallel.


froddd (approved):
Let's give this a go 👍

PR #1067, opened 2024-08-28
Tell cypress how to split up the tests between circle ci parallel runs.

aliuk2012:
# Related to
https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/1066

The previous PR configured the e2e job to run in parallel but didn't setup cypress to split up the tests for various runs


froddd (approved):
👍

PR #1075, opened 2024-09-02
Parallel run Github Actions for CI

aliuk2012:
# Context
Currently it takes ~15mins to run our CI Github Action everytime we push changes to a PR, Create a new PR and merge PR's. https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/actions/workflows/ci.yml?query=branch%3Amain

This is because we were running a single `script/test` script which was doing type checking, linting, unit & integration testing. As all these checks were done in sequence it meant that it would take 6-8mins before even starting the integration tests. It also meant that that if any of the steps failed you would have to scroll through the CI logs to identify the errors.

Its also not very clear which check is slow:

image

## Improvements
This PR reduces the time to run from ~15mins to ~10mins. This is done by running each check as separate job in parallel, i.e all the checks can start at exactly the same time instead of waiting for each other to first pass.  Its also clearer from the summary which step takes longs or has failed without having to scroll through the logs. From the screenshot below, you can see integration step(ie running cypress) takes that longest and we could speed that up by splitting up the task into a number of jobs using a matrix.

image

We should be able to remove "test" once this has been merged. Again you can see how long each step took to run.
image


.github/workflows/ci.yml
```diff
@@ -13,9 +13,59 @@ env:
...
+        uses: actions/setup-node@v4.0.3
+        with:
+          node-version-file: '.node-version'
+          cache: 'npm'
+
+      - name: Installing dependencies
+        run: npm install       
+
+      - name: Building source
+        run: npm run build
```
froddd: I don't think we need to run the full build to check linting or typechecking? This should speed things up yet a bit more (albeit for the fastest jobs anyway!)

.github/workflows/ci.yml
```diff
@@ -13,9 +13,59 @@ env:
...
+        uses: actions/setup-node@v4.0.3
+        with:
+          node-version-file: '.node-version'
+          cache: 'npm'
+
+      - name: Installing dependencies
+        run: npm install       
 
+      - name: Building source
+        run: npm run build
```
froddd: Same here -- no need to build?

.github/workflows/ci.yml
```diff
@@ -13,9 +13,59 @@ env:
...
+        run: npm install       
+
+      - name: Building source
+        run: npm run build
+     
+      - name: Running Lint checks
+        run: npm run lint
+    
+      - name: Pulling the latest type from the API repo
+        run: npm run generate-types
```
froddd, line 29: I still think this shouldn't be run on every test, as we have separate PRs for updating those types, so now feels like a good time to remove it

.github/workflows/ci.yml
```diff
@@ -24,17 +74,35 @@ jobs:
         uses: actions/setup-node@v4.0.3
         with:
           node-version-file: '.node-version'
+          cache: 'npm'
+
+      - name: Installing dependencies
+        run: npm install       
+
+      - name: Building source
+        run: npm run build
```
froddd: Again, we shouldn't need to build here

froddd (commented):
I think we can remove a few more steps to save on total resources used. 

Also is it worth replacing `npm install` with `npm ci` all around?

.github/workflows/ci.yml
```diff
@@ -13,9 +13,59 @@ env:
...
+     
+      - name: Running Lint checks
+        run: npm run lint
```
aliuk2012: No need to run this here as there is already a job for that

.github/workflows/ci.yml
aliuk2012: Lets keep this in for the moment, I think now that its separated into its own job we can easily see that the unit/integrations tests pass and until the types are updated a specific PR will be blocked. Its als osuper quick. We dont have to then mark a PR ready for review, get it approved and then have to put it back into draft so that it doesn't accidentally get merged

aliuk2012: Lets keep this in for the moment, I think now that its separated into its own job we can easily see that the unit/integrations tests pass and until the types are updated a specific PR will be blocked. Its als osuper quick. We dont have to then mark a PR ready for review, get it approved and then have to put it back into draft so that it doesn't accidentally get merged

froddd (approved):
🚀

williamfalconeruk (approved):
🏆

PR #1076, opened 2024-09-02
Ensure existing PDU is valid or flag section as 'In progress'

froddd:
# Context

https://dsdmoj.atlassian.net/browse/CAS-589

When an application was started before the change to the Practitioner details screen, old data would still pass validation and mark the step as 'Completed', when it should have been marked as 'In progress' instead, with subsequent steps showing as 'Cannot start yet'. This data would however trigger a validation error in the API, preventing the user from submitting the application.

This ensures that, when loading the task list, any old PDU data is considered invalid, and prevents the user from attempting to submit the application without updating the Contact details section.

### DIFF
--- a/wiremock/stubs/pdus.json
+++ b/wiremock/stubs/pdus.json
@@ -150,7 +150,7 @@
     "serviceScope": "temporary-accommodation"
   },
   {
-    "id": "f6fd8e4d-50a4-438d-83b8-335c538b8b66",
+    "id": "f6fd8e4d-50a4-438d-83b8-335c538b8b67",
     "name": "Essex South",
     "isActive": true,
     "serviceScope": "temporary-accommodation"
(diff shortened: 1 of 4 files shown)


wiremock/stubs/pdus.json
```diff
@@ -150,7 +150,7 @@
     "serviceScope": "temporary-accommodation"
   },
   {
-    "id": "f6fd8e4d-50a4-438d-83b8-335c538b8b66",
+    "id": "f6fd8e4d-50a4-438d-83b8-335c538b8b67",
```
froddd, line 153: This should solve this absolutely random test failure 😄  https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/actions/runs/10669075094/job/29570513513#step:7:553

aliuk2012: is there anyway we can make this human friendly reference instead of a "magic number"? It will make it a lot easier to spot issues like incorrect region/pdu.

froddd: These are stubs so replicate database data; we could theoretically use more human-friendly ids but then we're deviating from what the DB returns, which probably isn't desirable either? After this change, there are no longer any PDU with a duplicated id in this stub:

```js
const pdus = require('./wiremock/stubs/pdus.json')
pdus.map(x => x.id).length === [...new Set(pdus.map(x => x.id))].length // returns true
```

aliuk2012 (approved):

PR #1079, opened 2024-09-04
Add local headless E2E test command

froddd:
This consolidates commands under the `scripts` property of the package.json file, which is more discoverable for node projects than the scripts inside the `./script` directory, and more consistent with the location of other test scripts.


aliuk2012 (approved):
I like this approach, one small suggestion is that maybe we should add a mention in the README that ap-tools needs to be running before attempting to run these tasks because currently `script/local_e2e` starts up ap tools and alerts if it cant find the ap-tools command.

PR #1083, opened 2024-09-05
Cas 82/enable 0 day turnarounds

froddd:
# Context

https://dsdmoj.atlassian.net/browse/CAS-82

# Changes in this PR

Updates error messages for turnaround errors, and add hint to ensure value entered is zero or more.

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/server/utils/formUtils.ts
+++ b/server/utils/formUtils.ts
@@ -196,3 +196,13 @@ export const parseNaturalNumber = (input: string): number => {
   }
   return Number.parseInt(input, 10)
 }
+
+export const parseIntegerNumber = (input: string): number => {
+  if (!input) {
+    return undefined
+  }
+  if (!/^[-0-9]+$/.test(input)) {
+    return Number.NaN
+  }
+  return Number.parseInt(input, 10)
+}
--- a/server/utils/formUtils.ts
+++ b/server/utils/formUtils.ts
@@ -187,22 +187,32 @@ export function isStringOrArrayOfStrings(input: unknown) {
   )
 }
 
-export const parseNaturalNumber = (input: string): number => {
-  if (!input) {
-    return undefined
-  }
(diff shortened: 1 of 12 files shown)


server/utils/formUtils.ts
```diff
@@ -196,3 +196,13 @@ export const parseNaturalNumber = (input: string): number => {
...
+
+export const parseIntegerNumber = (input: string): number => {
+  if (!input) {
+    return undefined
+  }
+  if (!/^[-0-9]+$/.test(input)) {
+    return Number.NaN
+  }
+  return Number.parseInt(input, 10)
+}
```
froddd: I'm aware that adding this when there is a `parseNaturalNumber` method just above is odd, but I'd be a tad worried that updating the existing method may introduce bugs. Open to suggestions -- the existing one could take an extra argument (`options`?), but would need to be renamed across the project anyway (`parseNumber`?)!

aliuk2012: I agree it is a bit odd to have similar methods and to try play spot the diff with the regex. I like suggestion of a single `parseNumber` method. Perhaps rename the existing `parseNaturalNumber` to `parseNumber` and by default that  method would parse natural numbers. Once thats done extend the `parseNumber` to accept an optional argument which could be `options`. This should hopefully reduce the risk of introducing bugs.

This optional argument could be a json object for different types of numbers we might want to parse. Do we have any fields asking for monetary values? maybe something `options = { allow_negatives: true, allow_decimals: false }` . For the time being the only place where `options` would be set is the turnaround times.

aliuk2012 (approved):
Awesome refactor much better to extend the existing parseNumber method 👏

PR #1085, opened 2024-09-05
add demo build task to circle ci config [DO NOT MERGE!]

Rosie-Brigham:
# DO NOT MERGE!

# Context
This PR attempts to add logic into the circle ci config, so that and branch that starts with the word "demo" can be released (upon approval) onto our demo environment.

However, branch filtering locig is being a pain. I have created this PR to try and see if that will trigger the build. Please don't merge it yet!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -87,6 +87,15 @@ workflows:
   version: 2
   build-test-and-deploy:
     jobs:
+      - request-demo-build-approval:
+          type: approval
+          requires:
+            - helm_lint
+          filters:
+            branches:
+              only:
+                - /demo-.*/
+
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -136,6 +153,29 @@ workflows:
...
 
+      - request-demo-approval:
+          type: approval
+          requires:
+            - helm_lint
+            - build_docker_for_demo
+          filters:
+            branches:
+              only:
+                - /demo-.*/
```
froddd: Wonder if this is entirely necessary? I imagine one single approval step would be enough to trigger a build _and_ the deploy

PR #1088, opened 2024-09-10
CAS-399 - Add nav list to global navigation component

aliuk2012:
# Context

https://dsdmoj.atlassian.net/browse/CAS-399

# Changes in this PR
Any user who has `assessor` role will now see a list of links in the primary navigation. If they do not `assessor` role then they will not see any links in the primary navigation.

As part of this PR, there were changes to the bookings index page instead of `/bookings/{status}` we now have a `/bookings` that will redirect the user to the first page of bookings by default.

The url to access the reports page also changed from `/reports/bookings/` to `/reports`

### After

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod


e2e/tests/stepDefinitions/manage/bookingReport.ts
```diff
@@ -3,7 +3,7 @@ import { Given, Then } from '@badeball/cypress-cucumber-preprocessor'
 import { Cas3ReportType } from '@approved-premises/api'
 import Page from '../../../../cypress_shared/pages/page'
 import DashboardPage from '../../../../cypress_shared/pages/temporary-accommodation/dashboardPage'
-import ReportNewPage from '../../../../cypress_shared/pages/temporary-accommodation/manage/reportNew'
+import ReportNewPage from '../../../../cypress_shared/pages/temporary-accommodation/manage/reportIndex'
```
froddd: Can we get this one renamed too? Obviously will also need other usages in this file done also!

```suggestion
import ReportIndexPage from '../../../../cypress_shared/pages/temporary-accommodation/manage/reportIndex'
```

server/controllers/temporary-accommodation/manage/reportsController.ts
```diff
@@ -34,6 +34,12 @@ export default class ReportsController {
     }
   }
 
+  new(): RequestHandler {
+    return async (req: Request, res: Response) => {
+      return res.redirect(301, paths.reports.index({}))
+    }
+  }
```
froddd: It may be nice to think of a way of flagging this as deprecated -- maybe renaming the method would be sufficient, or creating a generic decorator that would handle the redirect? It seems we'll be needing this in a couple of places, and it would help any new dev understand the reason behind the redirect.

server/middleware/populateCurrentUser.ts
```diff
@@ -13,6 +13,9 @@ export default function populateCurrentUser(userService: UserService): RequestHa
         req.session.userDetails = userDetails
         req.session.probationRegion = userDetails.region
 
+        // add nav here
```
froddd: This comment doesn't add much, can be removed.

server/routes/temporary-accommodation/manage.ts
```diff
@@ -252,7 +252,7 @@ export default function routes(controllers: Controllers, services: Services, rou
     ],
   })
 
-  get(paths.reports.new.pattern, reportsController.new(), { auditEvent: 'VIEW_REPORT_CREATE' })
+  get(paths.reports.index(), reportsController.index(), { auditEvent: 'VIEW_REPORT_INDEX' })
```
froddd, line 260: We still need a reference to the current `paths.reports.new()` here, otherwise it won't be handled at all (as it stands trying to load `/reports/bookings` returns an error)

In fact this may be a better place to hold the redirect itself, with a new `redirectsController` maybe? This would remove the need to add the redirect logic everywhere it is needed, and possibly make all redirects more discoverable, and easier to implement when needed.

```suggestion
  // Redirect using a generic redirects controller, with redirect method taking location and optional status code
  get(paths.reports.new(), redirectsController.redirect(paths.reports.index(), 301), { auditEvent: 'REDIRECT_REPORT_NEW' })
  get(paths.reports.index(), reportsController.index(), { auditEvent: 'VIEW_REPORT_INDEX' })
```

server/routes/temporary-accommodation/manage.ts
```diff
@@ -302,6 +302,10 @@ export default function routes(controllers: Controllers, services: Services, rou
     ],
   })
 
+  get(paths.bookings.index(), bookingsController.index(), { auditEvent: 'VIEW_BOOKINGS_INDEX' })
+
+  get(paths.bookings.index(), bookingsController.index(), { auditEvent: 'VIEW_BOOKINGS_INDEX' })
```
froddd: Unnecessary duplication?

Could also be handled by a generic redirects controller as mentioned earlier?

server/services/userService.test.ts
```diff
@@ -52,4 +54,42 @@ describe('User service', () => {
...
+        { text: 'Reports', href: paths.reports.index({}) },
+      ]
+
+      it('returns a list of navigation items for the user', () => {
+        const result = userService.getActingUserPrimaryNavigationList(assessorUser, '')
+        expect(result).toEqual(navigationList)
+      })
+
+      describe('when a user is viewing a nav list item page', () => {
+        it.each(navigationList)(`returns active when a user is viewing %p page with %p link`, ({ text, href }) => {
```
froddd, line 91: Could this also test the correct behaviour with paths that only _contain_ the nav url? I.e. showing 'Bookings' as active for `/bookings/foo/bar`

server/services/userService.ts
```diff
@@ -30,4 +33,28 @@ export default class UserService {
...
+      ]
+
+      navList.every(item => {
+        if (currentPage.includes(item.href)) {
+          item.active = true
+          return false
+        }
+        return true
+      })
+      return navList
```
froddd: This feels odd -- `every` is typically used to assert every element meets a condition. Could this use `map` instead?

```suggestion
      return navList.map(item => ({
        ...item,
        active: currentPage.includes(item.href) || undefined, // or leave this as `false` instead of undefined
      }))
```

froddd (commented):
Some suggestion regarding handling of redirects to make them more discoverable.

It would also be good to have some integration test for the nav, to show that it is actually getting rendered correctly depending on the user. There are some integration tests around roles already, maybe add to that?

server/routes/temporary-accommodation/manage.ts
aliuk2012: Love this suggestion, I was following the existing pattern that was used for Referralsl. I'll update referrals as well so theres no redirects happening in the controllers

server/controllers/temporary-accommodation/manage/reportsController.ts
aliuk2012: I'm thinking we drop this completely in favour or doing the redirect directly in the routes file. Any thoughts?

froddd: Yes, that seems like the best option -- less code when new redirects are needed, and the concern of routing is all kept in one place. Better!

froddd (approved):
Perfect -- I think moving to the RedirectsController approach works really well. Keeps concern in one place which is really nice!

PR #1092, opened 2024-09-11
Ensure test search matches premises attributes

froddd:
Fix failing CAS3 E2E tests.

The bedspace search test was producing flaky results, as the attributes checked for said search were selected at random. In some cases, this matched the attributes of the premises created earlier in the test and used for context, but more often than not, there was no match, meaning no results were returned and a failing test.

This fix ensures the search attributes are selected based on the ones applied to the premises created earlier, which means the bedspace result should always be returned.


davidatkinsuk (approved):
LGTM 👍

muhammad-elabdulla (approved):

PR #1094, opened 2024-09-11
Skip testing empty PDU field creates error

froddd:
This should be temporary: until the mvoe to searching for multiple PDUs is complete, the API is unable to create errors for all fields at once, and instead returns errors for startDate and durationDays being required first, and an error for probationDeliveryUnit being required only if those two other fields are valid.

We can skip testing for the PDU error for now, but will reinstate it when the API only accepts a `probationDeliveryUnits` parameter as a list of PDUs.


muhammad-elabdulla (approved):

daveawc (approved):

muhammad-elabdulla:
@froddd can add this comment to the FE ticket

PR #1097, opened 2024-09-19
Trim user input strings

aliuk2012:
Repeat of previous config added to CAS 1 here https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/2137


froddd (approved):
👍

PR #1101, opened 2024-09-24
Fix failing e2e test not finding pdu in bedspace search page

aliuk2012:
Caused by https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/1099

PR #1099, opened 2024-09-23
CAS-467:  Update the bedspace search component to include multiple selection of pdu

aliuk2012:
# Context
https://dsdmoj.atlassian.net/browse/CAS-467

# Changes in this PR

### Before
image
image

### After
image
image

# Release checklist

[Release process documentation](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/edit-v2/4247847062?draftShareId=a1c360ab-bd31-4db1-aae3-7cc002761de9).

## Pre-merge checklist

- [ ] Are any changes required to dependent services for this change to work?
  (eg. CAS API)
- [ ] Have they been released to production already?

## Post-merge checklist

- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to test
- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to preprod
- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to prod

PR #1102, opened 2024-09-25
PDU id being incorrectly set to PDU name in e2e tests

aliuk2012:
# Context

In Circle CI, we were seeing e2e tests failing for the full apply-and-place feature. The tests were failing because they were looking for a PDU checkbox on the search bedspace page with a value relating to the PDU name. This was recently changes in #1099 instead of a single PDU name, we were passing a collection of PDU ids. We partially fixed the failing e2e for Bedspace search feature in #1101.  

It was odd that Bedspace search feature was passing yet apply-and-place feature was failing. There are two difference in the way the tests are setup.

Bedspace search feature generates pdus from factories where apply-and-place generate PDU's through the UI and scrap the screen to find the details needs for the models.


davidatkinsuk (approved):
LGTM 👍

PR #1103, opened 2024-09-26
Use the PDU ID that is on the premises show page

aliuk2012:
# Related

- #1102 
- #1101 
- #1099 

# Context

the apply-and-place feature tests were failing when the feature got to "Search for available bedspace" page because it could not find a checkbox with a pdu id from an existing premises. After debugging the tests it was discovered that we had set the premises id as the PDU id when creating the PDU from a factory. see #1102 

apply-andplace features use the ui to create the various assessments/premises etc and scrape the screen to get enough information to use with factories to store the data in memory for the tests to use dynamically. The problem with this approach is that the pdu id was not surfaced to the ui at all so could not use it when building the pdu.

I decided to surface the pdu id on the premises page where most of the information about the premises is surfaced and place it in a html data attribute called `data-cy-pdu-id`. The tests can now read that id from the page and store it for later use on the search available bedspaces page

### DIFF
--- a/cypress_shared/pages/temporary-accommodation/manage/premisesShow.ts
+++ b/cypress_shared/pages/temporary-accommodation/manage/premisesShow.ts
@@ -40,16 +40,20 @@ export default class PremisesShowPage extends Page {
                           .contains('Expected turnaround time')
                           .siblings('.govuk-summary-list__value')
                           .then(turnaroundTimeElement => {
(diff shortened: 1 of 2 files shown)


davidatkinsuk (approved):
LGTM 👍

froddd (approved):
Happy for this to go to unblock pipelines, but the other suggested approach should be the eventual aim!

PR #1143, opened 2024-11-18
Run e2e-tests locally against dev upstream services to simulate pipeline

gregkhawkins:
# Context

- @davidatkinsuk recently made changes to AP tools so that the AP tools local-api integrates with dev upstream services
- these new AP tools changes gives us the opportunity to locally run e2e tests against dev upstream services
- the advantage of this is that it allows us to locally simulate what happens in the CircleCI pipleine 
- this will make it easier to fix failures in the `cas3_e2e_tests` job 

# Changes in this PR
- new `npm scripts` to run e2e tests in this new way
- changes to the `README.md` to document this
- new `.env` config file for the new `npm scripts` to use

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,6 +129,31 @@ Spin up a real browser that gives you an interface to run individual E2E feature
 npm run test:e2e:ui
 ```
 
+#### End-to-End tests (via Cypress) using dev-upstream AP Tools
+This enables us to locally simulate the e2e test run in the Circle-CI pipeline
+
+First start the ap-tools using
+```
+ap-tools server stop --clear-databases
+ap-tools server start --local-ui --local-api-dev-upstream
+```
+
+* The end-to-end tests run against our local API which then integrates with the dev upstream services and the dev database due to this param `-local-api-dev-upstream`
+* Once AP is running successfully, we can run one of the two npm scripts below which both use the `e2e.dev-upstream.env` config file (instead of `e2e.local.env` config file)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -129,6 +129,31 @@ Spin up a real browser that gives you an interface to run individual E2E feature
...
+This enables us to locally simulate the e2e test run in the Circle-CI pipeline
+
+First start the ap-tools using
+```
+ap-tools server stop --clear-databases
+ap-tools server start --local-ui --local-api-dev-upstream
+```
+
+* The end-to-end tests run against our local API which then integrates with the dev upstream services and the dev database due to this param `-local-api-dev-upstream`
+* Once AP is running successfully, we can run one of the two npm scripts below which both use the `e2e.dev.env` config file (instead of `e2e.local.env` config file)
```
davidatkinsuk: can we please rename the config file e2e.dev-upstream.env for consistency with cas1?

e2e.dev.env
```diff
@@ -0,0 +1,7 @@
+CYPRESS_assessor_username=TEMPORARY-ACCOMMODATION-E2E-TESTER
+CYPRESS_assessor_password=secret
+CYPRESS_referrer_username=TEMPORARY-ACCOMMODATION-E2E-REFERRER
+CYPRESS_referrer_password=secret
```
davidatkinsuk: instead of 'secret' can we use 'doesntmatter' for consistency with cas1 config (and to indicate it's irrelevant :) )

davidatkinsuk (approved):
Thanks for doing this, it's going to be really helpful going forward 👍

PR #1162, opened 2024-12-08
CAS-603 Implement timeline for probation practitioners

aliuk2012:
# Context
https://dsdmoj.atlassian.net/browse/CAS-603

# Changes in this PR

Add timeline to PP "applications" so that they can see the progress of any events for the applications they submit. The timeline will exclude all HPT notes.

### Before
image

### After
image

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/server/data/timelineClient.test.ts
+++ b/server/data/timelineClient.test.ts
@@ -1,0 +1,48 @@
+import nock from 'nock'
+
+import { fakerEN_GB as faker } from '@faker-js/faker'
+import config from '../config'
+import paths from '../paths/api'
+import TimelineClient from './timelineClient'
+
+import { CallConfig } from './restClient'
+import { referralHistorySystemNoteFactory } from '../testutils/factories'
+
+describe('Timeline client', () => {
+  let fakeApprovedPremisesApi: nock.Scope
+  let timelineClient: TimelineClient
+
+  const callConfig = { token: 'some-token' } as CallConfig
+
+  beforeEach(() => {
+    config.apis.approvedPremises.url = 'http://localhost:8080'
+    fakeApprovedPremisesApi = nock(config.apis.approvedPremises.url)
+    timelineClient = new TimelineClient(callConfig)
+  })
+
+  afterEach(() => {
+    if (!nock.isDone()) {
+      nock.cleanAll()
+      throw new Error('Not all nock interceptors were used!')
+    }
+    nock.abortPendingRequests()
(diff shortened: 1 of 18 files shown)


server/data/timelineClient.test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+      fakeApprovedPremisesApi
+        .get(paths.assessments.timeline({ assessmentId: payload }))
+        .matchHeader('authorization', `Bearer ${callConfig.token}`)
+        .reply(201, results)
+
+      const result = await timelineClient.fetch(payload)
+
+      expect(result).toEqual(results)
+      expect(nock.isDone()).toBeTruthy()
```
Rosie-Brigham, line 45: Is this relevant if we have the `afterEach` block?

server/data/timelineClient.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    if (!nock.isDone()) {
+      nock.cleanAll()
+      throw new Error('Not all nock interceptors were used!')
+    }
+    nock.abortPendingRequests()
+    nock.cleanAll()
+  })
+
+  describe('fetch', () => {
+    it('returns timeline results', async () => {
```
Rosie-Brigham, line 33: Would it be overkill to have a test which reflects what happens when the application is not submitted?
(Or, is there no chance of this being called if that is the case?)

server/utils/assessments/timelineUtils.ts
```diff
@@ -0,0 +1,131 @@
+import {
```
Rosie-Brigham, line 1: Nice tidy up

Rosie-Brigham (approved):
Looks good - one comment but tine. Lovely bit of refactoring :)

server/data/timelineClient.test.ts
aliuk2012: The timelines at the moment, would only be included on pages where an application has been submitted and therefore it would have an assessmentId.

I think its a bit overkill for the moment, but if we were to place timelines on say the application pages I think we would have to look at how passing application id would work and at that point we could review scenarios where application is not submitted

server/data/timelineClient.test.ts
aliuk2012: I think its more explicit here that we want to make sure all mocked expectations have been done.

PR #1204, opened 2025-01-29
Fix flaky end-to-end tests for bedspace search

aliuk2012:
# Context
Issue introduced  as part of  https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/1153

Because we randomly select characteristics for premise and bedspace, iterating through them sometimes results in no searchable filters to return data for. Now we seperated out the wheelchair and occupancy characteristics, we now have to lookup the bedspace characteristic as well.


brimchugh (approved):
LGTM

PR #1209, opened 2025-02-04
Suppress part of the feature specs that are flakey

aliuk2012:
# Context
We are having an issue where the assessment is somes not on the page. We suspect this is due to an issu with the tests rather than the service itself. Its been tricky to debug as its so inconsistent and random. Instead of blocking the pipeline, we have decided to suppress the tests in question.

# Changes in this PR

### After

# Release checklist

[Release process documentation](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/edit-v2/4247847062?draftShareId=a1c360ab-bd31-4db1-aae3-7cc002761de9).

## Pre-merge checklist

- [ ] Are any changes required to dependent services for this change to work?
  (eg. CAS API)
- [ ] Have they been released to production already?

## Post-merge checklist

- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to test
- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to preprod
- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to prod

PR #1210, opened 2025-02-05
Fix failing specs

aliuk2012:
# Changes in this PR

Reverts 5 commits introduced as part of the following PR:

- https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/1153
- https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/1204
- https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/1206
- https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/1209

The reason for reverting these commits is that e2e tests seemed to be passing reliably before https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/1153 was merged.  Since it was merged CAS3 UI e2e and API CAS3 UI e2e tests have been failing for a number of reasons. Its been difficult to replicate/ debug locally because the tests do not also fail consistently.

These changes will mean that the API pipeline should not be blocked by the failing e2e tests anymore and it buys us some time to fix it. However, we will need to fix this before we do another CAS3 UI release otherwise the bedspace search page will revert to using checkboxes instead of radio buttons.

Brief summary of the failures we were seeing:

1. Failed to find a recently submitted application in the assessors list of unallocated referrals.

This is odd one as it doesn't seem to be linked to any of the above changes made to the bedspace search filters.

https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/fb293240f6eee0a0987aaa79a4e76bf6f223e018/e2e/tests/apply-and-place.feature#L17

2. Failed to find available bedspace

This is more likely to b caused by the changes as we made changes to the search filter and we suspect theres an issue when premises/bedspaces are created with random attributes. These random attributes could mean that the search results would not include the bedspace when we are expecting it too.

I've run the e2e tests locally multiple times on this branch and have yet to see any of the failures we were seeing before


Rosie-Brigham (approved):
Let's get this merged to unblock the pipeline

PR #1234, opened 2025-02-24
Configure Playwright for e2e tests (inc. dummy test with login)

gregkhawkins:
# Context

* `JIRA`: https://dsdmoj.atlassian.net/browse/CAS-1461
* `Do you need to add any environment variables?` --> Yes added to `.env-local` as you will see in the PR. Nothing needed for pipeline yet as out of scope of this ticket
* `Is an ADR required? An ADR should be added if this PR introduces a change to the architecture` --> There will need to be but out of scope of this ticket. Will be done as part of:  https://dsdmoj.atlassian.net/browse/CAS-1467

# Changes in this PR

The scope this ticket was to:
* Configure Playwright within this repo - mirror CAS1 Playwright configuration
* Add beginning of new test that runs against local AP tools - model on CAS1 Playwright standards
* New test to include; login into the web app successfully + ensured landed on the `Temporary Accommodation` web application's `Dashboard`

## Screenshots of UI changes
playwright

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod


e2e_playwright/playwright.config.ts
```diff
@@ -0,0 +1,33 @@
+import 'dotenv/config'
```
aliuk2012, line 1: This doesn't seem to be loading in the env vars for me automatically.  When I run the tests I'm getting `Error: Credentials missing, have you set the correct environment variables?`

e2e_playwright/playwright.config.ts
```diff
@@ -0,0 +1,33 @@
...
+    {
+      name: 'setup-local',
+      testMatch: /.*\.setup\.ts/,
+      use: { baseURL: 'http://localhost:3000' },
+    },
+    {
+      name: 'local',
+      use: {
+        ...devices['Desktop Chrome'],
+        baseURL: 'http://localhost:3000',
```
aliuk2012: Just a thought, instead of hardcoding the url here, could we maybe use env vars? It would make it more flexible should we want to run the local server on other ports or run it against dev/test/pre-prod etc.

I know theres a separate bit of config that needs to be added once we start running this against dev so perhaps we can address it then. I just dont like all the hard coding of urls for config, single source of truth 😉

e2e_playwright/playwright.config.ts
gregkhawkins: From memory, I had to move `dotenv` in `package.json` from `dev-dependencies` -> `dependencies`. Might be worth re-running `npm install`

e2e_playwright/playwright.config.ts
gregkhawkins: will do, thanks!

e2e_playwright/playwright.config.ts
gregkhawkins: also, assume you copied the new env vars from `.env.local` into your `.env` file already?

e2e_playwright/playwright.config.ts
gregkhawkins: done :)

e2e_playwright/playwright.config.ts
aliuk2012: 🤦‍♂️  I hadn't copied the new vars from `.env.local` to `.env`

aliuk2012 (approved):

PR #1246, opened 2025-03-06
CAS-1482: Move bedspace search api endpoint to cas 3

aliuk2012:
# Context
https://dsdmoj.atlassian.net/browse/CAS-1482

# Changes in this PR

We need to switch over to a new bedspace search endpoint that is now cas3 namespaced. This means that we no longer have to pass the service name when calling it and its more obvious than the old /beds/search.  There were also a few types that had to be updated as result of the changing the endpoint point but nothing that would change functionality for the user.

We dont use the concepts of Beds in CAS3 so I thought it would be a good opportunity to address some tech debt and to rename any thing that relates to beds to bedspace.


cypress_shared/components/bedspaceSearchResult.ts
```diff
@@ -1,10 +1,10 @@
-import { TemporaryAccommodationBedSearchResult as BedSearchResult } from '../../server/@types/shared'
```
Rosie-Brigham, line 1: ❤️

Rosie-Brigham (approved):
Looks good to me - lovely bit of refactoring
As a small caveat I wasn't able to look too much into detail as looked at this between meetings

PR #1249, opened 2025-03-07
Playwright CI config changes

gregkhawkins:
# Context
* The `cas3 e2e job` was failing in the API pipeline
* Strangely the same e2e tests  passed in the `hmpps-temporary-accommodation-ui` pipeline
* Locally the same e2e tests were consistently passing too

# Changes in this PR
* It turned out that setting the shards in the way I was with this config `npm run test:playwright:e2e:ci -- --shard=${SHARD}/${CIRCLE_NODE_TOTAL}` means that the `hmpps-temporary-accommodation-ui` pipeline was only running 2 out of 6 tests - this is why this pipeline was passing
* The reason the pipeline consistently passes locally but not in the CI pipelines is still a bit of a mystery tbh
* However, the reason for the pipeline failures was that using `this.page.getByLabel('Start date').fill(startDate)` was flaky and did not always set the date value
* To fix this so that it works consistently I have used the datepicker to set the date. This solution works for locally and for both pipelines and gives us more coverage on the page - so win win :)


.circleci/config.yml
```diff
@@ -85,15 +85,14 @@ jobs:
       - run:
           name: Run E2E tests
           command: |
-            SHARD="$((${CIRCLE_NODE_INDEX}+1))"
             export ASSESSOR_USERNAME=${CYPRESS_ASSESSOR_USERNAME_<< parameters.environment >>} ASSESSOR_PASSWORD=${CYPRESS_ASSESSOR_PASSWORD_<< parameters.environment >>} REFERRER_USERNAME=${CYPRESS_REFERRER_USERNAME_<< parameters.environment >>} REFERRER_PASSWORD=${CYPRESS_REFERRER_PASSWORD_<< parameters.environment >>} ACTING_USER_PROBATION_REGION_ID=${CYPRESS_ACTING_USER_PROBATION_REGION_ID_<< parameters.environment >>} ACTING_USER_PROBATION_REGION_NAME=${ACTING_USER_PROBATION_REGION_NAME_<< parameters.environment >>} ENVIRONMENT=${CYPRESS_ENVIRONMENT_<< parameters.environment >>} DEV_PLAYWRIGHT_BASE_URL=https://transitional-accommodation-<< parameters.environment >>.hmpps.service.justice.gov.uk
-            npm run test:playwright:e2e:ci -- --shard=${SHARD}/${CIRCLE_NODE_TOTAL}
+            npm run test:playwright:e2e:ci
```
gregkhawkins, line 89: It turned out that settings the shards in this way lead to the pipeline only running 2 / 6 tests. Removing this here means we now run all tests

.circleci/config.yml
```diff
@@ -85,15 +85,14 @@ jobs:
...
       - store_artifacts:
-          path: playwright-report
-          destination: playwright-report
+          path: e2e_playwright/playwright-report
+          destination: e2e_playwright/playwright-report
       - store_artifacts:
-          path: test-results
-          destination: test-results
+          path: e2e_playwright/test-results
+          destination: e2e_playwright/test-results
```
gregkhawkins, line 95: the artifacts were not available in the artifacts tab of the build before making this change. We can now watch the failure videos in the pipeline when failures happen so easier to diagnose issues

.env.local-dev-upstream
```diff
@@ -1,4 +1,4 @@
-ENVIRONMENT=dev
+ENVIRONMENT=local
```
gregkhawkins, line 1: realised that `ap-tools server start --local-ui --local-api-dev-upstream` was broken. This change fixed it

e2e_playwright/pages/reports/reportsPage.ts
```diff
@@ -2,27 +2,35 @@ import { Page, expect } from '@playwright/test'
...
+    await this.pickStartDateUsingDatePicker()
+    await this.pickEndDateUsingDatePicker()
   }
 
   private async selectProbationRegion(probationRegionName: string) {
     await this.page.getByLabel('Select a probation region').selectOption(probationRegionName)
   }
 
-  private async enterStartDate(startDate: string) {
-    await this.page.getByLabel('Start date').fill(startDate)
```
gregkhawkins, line 22: weirdly this was flaky - think this it's due to the target input field being part of the date-picker but bit of a mystery tbh. 
Anyway, below changes pick the date using thr date-picker component which gives us more coverage (and consistently works in the pipeline)

aliuk2012: Good catch

aliuk2012 (approved):

PR #1261, opened 2025-03-20
Playwright - Premise Management e2e tests

gregkhawkins:
# Context

JIRA: https://dsdmoj.atlassian.net/browse/CAS-1508

# Changes in this PR
- Coverage in playwright to replace the `Manage Premises` e2e tests in `premises.feature`

## Screenshots of UI changes
playwright_tests

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod


e2e_playwright/test.ts
```diff
@@ -20,4 +21,18 @@ export const test = base.extend<TestOptions>({
...
+    town: faker.location.city(),
+    postcode: faker.location.zipCode(),
+    notes: faker.lorem.lines(5),
+    turnaroundWorkingDayCount: faker.number.int({ min: 1, max: 5 }),
+    localAuthority: undefined,
+    probationRegion: undefined,
+    pdu: undefined,
+    propertyAttributesValues: [],
+    status: undefined,
+  },
```
gregkhawkins, line 37: I created this `property` test data here to avoid creating things in factories (and tieing the test data to the production code). Think you suggested avoiding using the factories way back when @aliuk2012 so think we are aligned on this? 
Did however, think it wise to use the fakers (as i have above) to generate random values. The `undefined` values above are a combination of values set by enums and reference-data (from the BE and DB). I left these values as `undefined` here and subsequently pick values at random from the selects and checkboxes on the UI which i then swap out for the underfined values. At the end of which we then have a object to assert against. This approach (picking values at random) should prevent future flakiness when there are reference-data changes in the DB etc.

I did wonder whether this is the best place to generate the `property` test object but I quite like it here instead of generating within the test (e.g. could alternatively create a `stubProperty()` function in the test itself). But this way it is global for reuse and felt aligned with the approach in CAS1 here: https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/e75945a4fb9285fca40865c5c658dd80baf4c057/e2e/test.ts#L5

Flexible on all of this though so let me know any concerns please :)

e2e_playwright/pages/manage/addPropertyPage.ts
```diff
@@ -0,0 +1,46 @@
...
+    property.pdu = await this.selectOptionAtRandom('What is the PDU?', 'Select a PDU')
+    property.propertyAttributesValues = await this.check3CharacteristicsAtRandom()
+    property.status = await this.clickRandomStatusRadioButton()
+
+    await this.page.getByLabel('Please provide any further property details').fill(property.notes)
+    await this.page
+      .getByLabel(
+        'Enter the number of working days required to turnaround the property. The standard turnaround time should be 2 days',
+      )
+      .fill(String(property.turnaroundWorkingDayCount))
```
aliuk2012, line 33: I think this is better off being tested as a Unit/integration test for the controller to ensure its handling validation errors

e2e_playwright/test.ts
aliuk2012: I like this approach and is much better than using the factories to drive the UI. I'm not opposed to using faker to create random data for this either. I think we were just over using factories when it came to integration/e2e testing.

gregkhawkins: excellent, thanks

e2e_playwright/pages/manage/addPropertyPage.ts
gregkhawkins: I think we need to enter the `turnaroundWorkingDayCount` so that we can submit the `Create property` form and make sure we get the success banner to test fully integrated with the API and successful. This is also the pre-requisite in both tests to doing other things (e.g. edit / save and check edited successfully).

I do agree though that checking the `Create property` form validation is working correctly does not belong here (which don't think it is). I think we have coverage for this form validation in this integration test:
https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/4c2d0bf6780be5aabb43722df5ff444708f2c232/integration_tests/tests/temporary-accommodation/manage/premises.cy.ts#L199

Let me know if there is extra coverage needed (or if I'm misunderstanding).

aliuk2012: 🤦‍♂️  On second reviewing this, I realised that the copy for label of the form field was leading me to think that we were checking a validation message and then filling in that field with number of days.  Label copy does seem rather lengthly and a bit like hint text.

Not an issue we can resolve here so going to  resolve this conversation

aliuk2012 (approved):

PR #1263, opened 2025-04-03
Migrate from CircleCI to GH Actions for security scans

aliuk2012:
### Related
- API - https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/3285
- CAS1 - https://github.com/ministryofjustice/hmpps-approved-premises-ui/pull/2442
- CAS2 - https://github.com/ministryofjustice/hmpps-community-accommodation-tier-2-ui/pull/904
- CAS3 - https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/1263

Result of running
https://github.com/ministryofjustice/hmpps-github-actions/blob/main/docs/workflow-migration.md#automation


.github/workflows/security_npm_dependency.yml
```diff
@@ -0,0 +1,12 @@
+name: Security npm dependency check
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "22 9 * * MON-FRI" # Every weekday at 09:22 UTC
```
froddd: Could we manually edit these times so all the checks happen before 8am for all CASs? I can see why we might want them staggered (that would be a whole lot of alerts flooding Slack and other channels at once otherwise!) but it probably would be nice to be able to start the work day with all needed alerts ready to be consulted :)

froddd (approved):
LGTM 👍

PR #1265, opened 2025-04-08
Setup Changes to accommodate new version of AP tools

gregkhawkins:
# Context
* Generate `.env` file using new `.env.template` file
* Generate `e2e.local.env` (for `Cypress`) file using new `e2e.local.env.template` file
* Pull secrets down from k8s to swap out secret values in `.env` and `e2e.local.env`
* e2es now run against latest version of `AP tools` (that by default integrates with upstream dev services)
* includes hack for running `Cypress` e2es locally to mitigate auth issue (commented in detail in this PR) in `e2e/tests/stepDefinitions/manage.ts` file
* removed all local e2e test data as no longer needed (as we are now pointing at `AP tools` that always integrates with dev upstream services so local data becomes irrelevant)


e2e/tests/stepDefinitions/apply.ts
```diff
@@ -25,14 +22,12 @@ Given('I start a new application', () => {
...
-    const pdu = {
-      id: '81ec3d88-eecd-49d7-ad90-5dff98080dfb',
-      name: 'East Sussex (includes Brighton and Hove)',
-    }
-    application.data['contact-details']['practitioner-pdu'] = pdu
-    application.data['contact-details']['probation-practitioner'].pdu = pdu
+  const pdu = {
+    id: '3fa6a349-5f48-4c42-8d96-938bda727c35',
+    name: 'Suffolk',
   }
```
gregkhawkins, line 28: this has changed because the e2e user is based in `East of England` (instead of current pipeline e2e user in `Kent, Surrey & Sussex`). The PDU needs to be within `East of England` or some of the e2e tests break.

Before we merge, to ensure we do not break the pipeline, we will need to change the e2e users in the pipeline (in the CircleCI pipeline secrets for UI and API) to match the dev users we are now using here. As a reminder we could not use the e2e users used by the pipeline because we do not know the passwords. Changing the e2e users in the pipeline is advisable anyway, so that what devs run locally exactly simulates what happens in the pipeline so that if it passes locally we are guaranteed it will pass in the pipeline.

e2e/tests/stepDefinitions/manage.ts
```diff
@@ -25,6 +27,11 @@ const signIn = (usernameVariable: string, passwordVariable: string) => {
...
   cy.get('input[name="username"]').type(username)
   cy.get('input[name="password"]').type(password, { log: false })
-
   cy.get('.govuk-button').contains('Sign in').click()
+
+  // temp hack to get local e2es working against dev upstream services - this will only be executed when e2es run locally (i.e. not in the pipeline)
+  if (environment === 'local') {
+    cy.wait(200)
+    cy.visit('/dashboard')
+  }
```
gregkhawkins, line 37: This was the hack I talked about in `Decouple AP tools - catchup`. This code ensures that this hack will only be used when running `Cypress` e2e tests locally. This is ensured with `if (environment === 'local')` and so the pipeline will continue to run the tests without this.

## Context
1. The deployment of the `CAS3` application (inc. loading of all `.env` secrets) is now handled outside the project by running `ap-tools server start --local-ui --local-api` - this is wok done as part of the `Decouple AP tools` firebreak work.
2. This means that the deployed app on http://localhost:3000 is exactly the same regardless of whether you are running e2e tests in `Cypress` or in `Playwright` - both use a `baseUrl` value of http://localhost:3000 so they both run tests against the local deployment and the behaviour should be the same regardless of e2e tech.
3. When we run e2es in `Playwright` notice that login is successfully and redirects to `Dashboard` as expected:
02_playwright_auth_success

4. When we run e2es in `Cypress` notice that login is not successful and redirects to `Authorization Error` page:
01_cypress_auth_error

N.B Notice above, in both screenshots, that we login as the same user. Therefore the  authentication/authorisation behaviour should be the same. Weirdly it is not!

## Findings
* It is not possible to re-create the `Authorization Error` we see with `Cypress` with the browser. When i login manually with Chrome with the same user, the brwser behaves in the same way as `Playwright` (i.e. redirects successfully to `Dashboard`)
* With `Cypress`, despite getting the `Authorization Error`, if you then go to the `Dashboard` (as the code above does when you are running with `environment = 'local'`), the application let's you through to the `Dashboard` (so you are authorized for this!) and in fact you can do everything that the e2e tests expect so you are 100% authorised.
* As a further attempt to understand what is going on with `Cypress` when this happens, I did manage to make these changes in the `setUpAuthentication.ts` file:
Screenshot 2025-04-09 at 16 51 00

* With the above changes, the result (when `Cypress` does a login) is that the `frontend` service in tilt logs this: 
```
fail message = Unable to verify authorization request state.
```
* This would suggest that the `state` parameter's value (that the Identity provider -  IDP - `hmpps-auth` I believe sends back to us) is invalid for some reason. 
* If this is happening, it should happen for all clients (i.e. `Chrome/Playwright and Cyprus`) and not only behave this way on `Cypress` login. Baffling!

## Decision
* Put this hack in as will only be applied when running e2e tests locally 
* We care less about testing the authentication locally (as auth managed owned outside this team and our integration never changes)
* As the hack is not applied in when running `environment = 'dev'` then if auth is genuinely broken the pipeline will still pick this up and e2es fail. 
* Given we used to entirely mock authentication when we ran e2es locally, I think we are good with this.
* Another thing to consider is that `Cyprus` e2es are going away soon - we are currently migrating everything over to `Playwright` so think we can live with this hack in the interim :)

PR #1267, opened 2025-04-09
Remove /applications/{applicationId}/documents endpoint

aliuk2012:
This is not used in CAS3, only cas1. I think this may have been copied over from cas1 when cas3 first forked.


froddd:
CAS3 no longer uses non-namespaced endpoints, so this is no longer needed.

PR #1277, opened 2025-04-17
Add Slack notification for E2E failure

froddd:
This adds a reusable action which can be used to notify of failure for any job. The actions from the HMPPS Github Actions repository are a bit over the top, so we've created our own.

Actions can not be run from branches and are only accessible when they are on the `main` branch, hence the need to merge this prior to testing the new Github Actions-based pipeline.


davidatkinsuk (approved):
very nice!

PR #1278, opened 2025-04-17
Update Slack notification action

froddd:
- Add optional `job` input to be used when the failing job is different than the one calling the action (useful for matrix job)
- Add the `text` property to payload (used for notifications)
- Prevent link unfurling in resulting message

An example message resulting from the action can be viewed here: https://mojdt.slack.com/archives/C08NHM3MA9K/p1744885251693359


davidatkinsuk (approved):
LGTM 👍

PR #1279, opened 2025-04-17
Chore/fix new ap tools integration

froddd:
# Changes in this PR

This PR reinstates the changes that had been introduced then reverted after the AP Tools improvements.

It updates the E2E tests (both Cypress and Playwright) to use the same user as all other environments: new environment variables have been added to the CircleCI config to match the users in all other envs. The variables related to the test user's region are now hardcoded for both E2E test suites.

The following pipeline on CircleCI shows the E2E pipeline tests passing with this configuration (CircleCI config was changed to run on this branch -- this commit has been removed): https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-temporary-accommodation-ui/8904/workflows/78ba8718-0390-4b88-9acf-bd1a1d70a02e

E2E tests are also passing locally. No screenshots, might post some later 😬 

_**Note:** given the first commit is a revert of a revert, it may be easier to review commits one by one!_

## Screenshots of UI changes

n/a


.circleci/config.yml
```diff
@@ -80,7 +80,7 @@ jobs:
       - run:
           name: Run E2E tests
           command: |
-            export ASSESSOR_USERNAME=${CYPRESS_ASSESSOR_USERNAME_<< parameters.environment >>} ASSESSOR_PASSWORD=${CYPRESS_ASSESSOR_PASSWORD_<< parameters.environment >>} REFERRER_USERNAME=${CYPRESS_REFERRER_USERNAME_<< parameters.environment >>} REFERRER_PASSWORD=${CYPRESS_REFERRER_PASSWORD_<< parameters.environment >>} ACTING_USER_PROBATION_REGION_ID=${CYPRESS_ACTING_USER_PROBATION_REGION_ID_<< parameters.environment >>} ACTING_USER_PROBATION_REGION_NAME=${ACTING_USER_PROBATION_REGION_NAME_<< parameters.environment >>} ENVIRONMENT=${CYPRESS_ENVIRONMENT_<< parameters.environment >>} DEV_PLAYWRIGHT_BASE_URL=https://transitional-accommodation-<< parameters.environment >>.hmpps.service.justice.gov.uk
+            export ASSESSOR_USERNAME=${E2E_USER_CAS3_ASSESSOR_USERNAME} ASSESSOR_PASSWORD=${E2E_USER_CAS3_ASSESSOR_PASSWORD} REFERRER_USERNAME=${E2E_USER_CAS3_REFERRER_USERNAME} REFERRER_PASSWORD=${E2E_USER_CAS3_REFERRER_PASSWORD} ENVIRONMENT=${CYPRESS_ENVIRONMENT_<< parameters.environment >>} DEV_PLAYWRIGHT_BASE_URL=https://transitional-accommodation-<< parameters.environment >>.hmpps.service.justice.gov.uk
```
gregkhawkins, line 83: nice, thank you! Presumably the `E2E_USER_CAS3_ASSESSOR_USERNAME` is the same as `CAS3_ASSESSOR_USERNAME` in k8s? Good that we are consistent between running locally nd pipeline

cypress.config.e2e.ts
```diff
@@ -46,6 +46,11 @@ export default defineConfig({
     setupNodeEvents,
     supportFile: false,
   },
+  env: {
+    // Probation region for user temporary-accommodation-training-1
+    acting_user_probation_region_id: 'ca979718-b15d-4318-9944-69aaff281cad',
+    acting_user_probation_region_name: 'East of England',
```
gregkhawkins, line 52: like it!

gregkhawkins (approved):

.circleci/config.yml
froddd: Yes, and I believe these are now kept in sync between k8s and the repositories thanks to David's magic script!

cypress.config.e2e.ts
froddd: Hardcoding these started from a red herring situation, but they're not actually secrets so it made sense to handle them this way instead of using environment variables.

gregkhawkins: yes indeed, nothing secret about them

PR #1280, opened 2025-04-22
Fix for flaky e2e

gregkhawkins:
**Context**
* We were getting some strange behaviour in the `Add and Edit property` playwright e2e test
* This happens when it clicks 3 characteristics when adding the property and we save the ids of the charcteristics nd assert on them in the edit screen
* The problem occurs when instead of clicking unique checkboxes, there is cross over so we check and then uncheck the same one
* Temporary work around is to only check one checkbox so there is no cross-over (will come back to this soon!)


froddd (approved):
LGTM

PR #1294, opened 2025-05-20
Change log-level to info + stop logging username

gregkhawkins:
# Context

* JIRA: https://dsdmoj.atlassian.net/browse/CAS-1653
* Discovered that we were logging usernames (which in most cases are personal emails)

# Changes in this PR
* Change the log level to `INFO` (from `DEBUG`) to stop the tokenVerifier logging this whenever token verification happens
* Log `user.id` in the error logs where we were previously logging username


williamfalconeruk (dismissed):

froddd (approved):
LGTM

PR #1298, opened 2025-05-22
Feature/cas 1521 date validation corrections

SaintRhys:
# Context

This addresses some comments on [this](https://dsdmoj.atlassian.net/browse/CAS-1521) ticket.

- Update validation messages.
- Add date validation on edit date arrival.

Also updates a move on category id that was causing an e2e test to randomly fail.

# Changes in this PR

### After

image

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/server/controllers/temporary-accommodation/manage/arrivalsController.ts
+++ b/server/controllers/temporary-accommodation/manage/arrivalsController.ts
@@ -4,7 +4,7 @@ import type { NewCas3Arrival as NewArrival } from '@approved-premises/api'
 import paths from '../../../paths/temporary-accommodation/manage'
 import { ArrivalService, BedspaceService, BookingService, PremisesService } from '../../../services'
 import { generateConflictBespokeError } from '../../../utils/bookingUtils'
-import { DateFormats, dateIsInFuture } from '../../../utils/dateUtils'
+import { DateFormats, dateIsInFuture, dateIsWithinLastSevenDays } from '../../../utils/dateUtils'
 import extractCallConfig from '../../../utils/restUtils'
 import {
   catchValidationErrorOrPropogate,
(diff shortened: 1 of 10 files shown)


server/controllers/temporary-accommodation/manage/arrivalsController.ts
```diff
@@ -153,6 +153,17 @@ export default class ArrivalsController {
           throw error
         }
 
+        if (updateArrival.arrivalDate) {
+          const sevenDays = new Date()
+          sevenDays.setDate(sevenDays.getDate() - 7)
```
gregkhawkins: we should probably use `date-fns` here like you have in the test. So for example:
```
const sevenDaysAgo = subDays(new Date(), 7)
```

gregkhawkins: think we have a little duplication in the `create()` function above.

gregkhawkins (changes requested):
Hiya, wondering whether we need to move some of the validation work you diud for this ticket to some validation utils files. For example, I created this when i was in CAS1:
https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/main/server/utils/match/validateSpaceBooking.ts

As flagged i noticed a little duplication and I had a quick look at the previous PR that Rob approved and can see that work might benefit from a shared validation utils also. Might be worth chatting it through?

server/utils/dateUtils.ts
```diff
@@ -225,3 +225,10 @@ export const dateIsWithinThreeMonths = (dateString: string): boolean => {
   const threeMonths = subMonths(dateString, 3)
   return isAfter(new Date(), threeMonths)
 }
+
+export function dateIsWithinLastSevenDays(dateString: string): boolean {
```
gregkhawkins, line 230: prob need a unit test for this one `dateUtils.test.ts`.

gregkhawkins:
> Hiya, wondering whether we need to move some of the validation work you diud for this ticket to some validation utils files. For example, I created this when i was in CAS1: https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/main/server/utils/match/validateSpaceBooking.ts
> 
> As flagged i noticed a little duplication and I had a quick look at the previous PR that Rob approved and can see that work might benefit from a shared validation utils also. Might be worth chatting it through?

gregkhawkins (approved):

PR #1319, opened 2025-06-24
Test/update generate types to pull from dev

daveawc:


script/generate-types
```diff
@@ -6,10 +6,15 @@ set -e
 
 cd "$(dirname "$0")/.."
 
-echo "==> Generating types..."
+if [ -z "$1" ]; then
+  environment="https://raw.githubusercontent.com/ministryofjustice/hmpps-approved-premises-api/main/src/main/resources/static/codegen/built-cas3-api-spec.yml"
```
gregkhawkins: does this not mean we no loinger build the shared models in `built-api-spec.yml`?

.github/workflows/generate-types.yml
```diff
@@ -1,7 +1,7 @@
 name: Generate Types
 
 on:
-  workflow_dispatch:
+  workflow_call:
```
froddd, line 5: Keep `workflow_dispatch` available here:

```suggestion
  workflow_call:
  workflow_dispatch:
```

script/generate-types
froddd: Maybe `environment` should be `apiSpec`?

script/generate-types
```diff
@@ -6,10 +6,15 @@ set -e
 
 cd "$(dirname "$0")/.."
 
-echo "==> Generating types..."
+if [ -z "$1" ]; then
```
daveawc, line 9: if the url isn't passed in as a parameter, then use the previous url (just whilst testing - will remove this in the future)

integration_tests/mockApis/assessments.ts
froddd, line 1: Not sure this should be included in this PR?

froddd (changes requested):
The changes to the workflow and script look fine, but I don't think any of the other files should have been touched!

gregkhawkins (approved):

PR #1322, opened 2025-06-25
Feature/cas 1711 display online properties tab summary counts

AvraamK:
# Context

Implements default v2 properties route (online) along with properties and bedspaces counts for v2 properties page. Also adds Archived Tab. 

More info here: 
https://dsdmoj.atlassian.net/browse/CAS-1711 and 
https://dsdmoj.atlassian.net/browse/CAS-1716

# Changes in this PR

- Auto-redirect to `v2/properties/online` when accessing `/v2/properties`
- Added new Archive tab in the v2 properties route
- Added counts for v2 online properties and their bedspaces above the results table

### Before
Screenshot 2025-06-25 at 12 42 50

### After
Screenshot 2025-06-26 at 20 01 36

# Release checklist

## Pre-merge checklist

  (eg. CAS API) 

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/server/controllers/temporary-accommodation/manage/v2/premisesController.test.ts
+++ b/server/controllers/temporary-accommodation/manage/v2/premisesController.test.ts
@@ -2,6 +2,7 @@ import { DeepMocked, createMock } from '@golevelup/ts-jest'
 import type { NextFunction, Request, Response } from 'express'
 
 import type { PremisesSearchParameters } from '@approved-premises/ui'
+import type { Cas3PremisesSearchResult } from '@approved-premises/api'
 import { CallConfig } from '../../../../data/restClient'
 import PremisesService from '../../../../services/v2/premisesService'
 import { probationRegionFactory } from '../../../../testutils/factories'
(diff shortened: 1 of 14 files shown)


robanderson-moj (changes requested):
The designs in Miro show the search term in the online properties count if the use has searched for an address

image



Also, it differs from the designs again when there are 0 results, and seems to remove the search box

Should be
image

Currently
image

server/controllers/temporary-accommodation/manage/v2/premisesController.ts
```diff
@@ -11,10 +13,25 @@ export default class PremisesController {
...
 
-      const params = req.query as PremisesSearchParameters
+      const params = req.query as PremisesSearchParameters & { status?: Cas3PremisesStatus }
 
-      const tableRows = await this.premisesService.tableRows(callConfig, params)
-      return res.render('temporary-accommodation/v2/premises/index', { tableRows, params })
+      // If no status parameter, redirect to include status=online
+      if (!params.status) {
+        const queryString = new URLSearchParams({ ...params, status: 'online' }).toString()
+        return res.redirect(`${paths.premises.v2.index({})}?${queryString}`)
```
gregkhawkins: what's the advantage of the redirect here? Can we just defult to `online` like this:
```
let { status } = params
if (!status) {
  status = 'online'
}
```

server/services/v2/premisesService.ts
```diff
@@ -7,21 +12,36 @@ import { CallConfig } from '../../data/restClient'
...
+            ]
+          })
+
+    return {
+      ...premises,
+      tableRows,
+    }
+  }
+
+  async tableRows(callConfig: CallConfig, params: PremisesSearchParameters): Promise<Array<TableRow>> {
```
gregkhawkins: as this is one is now only called by tests, perhaps this should just be:
```
async tableRows(premises: Cas3PremisesSearchResults): Promise<Array<TableRow>> {
    return premises.results === undefined
      ? []
      : premises.results.map(entry => {
          return [
            this.htmlValue(this.formatAddress(entry)),
            this.htmlValue(this.formatBedspaces(entry)),
            this.textValue(entry.pdu),
            this.htmlValue(`<a href="#">Manage</a>`),
          ]
        })
  }
```

The the test refactored to just test how were are mapping to the table data

server/services/v2/premisesService.ts
```diff
@@ -7,21 +12,36 @@ import { CallConfig } from '../../data/restClient'
 export default class PremisesService {
   constructor(protected readonly premisesClientFactory: RestClientBuilder<PremisesClient>) {}
 
-  async tableRows(callConfig: CallConfig, params: PremisesSearchParameters): Promise<Array<TableRow>> {
+  async searchData(
```
gregkhawkins: no unit test currently

server/services/v2/premisesService.ts
gregkhawkins: then swe call call this from `searchData()` function above and write the unit test. Or alternatively this can be moved to the controller:
```
const premisesClient = this.premisesClientFactory(callConfig)
const premises = await premisesClient.search(params.postcodeOrAddress ?? '', 'online')
```

Which would mean no need for the `searchData()` function

server/views/temporary-accommodation/v2/premises/index.njk
```diff
@@ -7,7 +7,7 @@
 
 {% extends "../../../partials/layout.njk" %}
 
-{% set pageTitle = "Online properties - " + applicationName %}
+{% set pageTitle = (isOnlineTab and "Online properties" or "Archived properties") + " - " + applicationName %}
```
gregkhawkins: we will need integration tests to cover the toggling of `Online properties` and `Archived properties` tabs. Not sure if e2e tests needed also?

gregkhawkins (changes requested):

server/views/temporary-accommodation/v2/premises/index.njk
gregkhawkins: also, had a quick look at other examples across `CAS3 UI` where we do similar stuff and found this:

Screenshot 2025-06-25 at 15 50 46 (2)

You will see that here the `Provisional bookings`, `Confirmed bookings`, `Active bookings` and `Departed bookings` tabs all have their own routes:
1. `Provisional bookings` = `/bookings/provisional`
2. `Confirmed bookings` = `/bookings/confirmed`
3. `Active bookings` = `/bookings/active`
4. `Departed bookings` = `/bookings/departed`

So think we probably want to drive the tabs with 2 routes:
1. `Online properties` - `/v2/properties/online`
2. `Archived properties` - `/v2/properties/archived`

From a quick look at how the bookings tabs are implemented looks like they have a shared template also so think your on the right lines with that. Other option is shared partials but let's not re-invent the wheel

AvraamK: https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/1322#discussion_r2166799133

I was thinking to add these new integration tests in the next ticket where I will add the `Archived properties` in order to fully test different tabs

gregkhawkins: as a rule of thumb.... we need test coverage for all impl changes in the PR. Otherwise can lose track of what is tested and what is not.

Perhaps this is a reason to make this PR the counts functionality only and can add int test coverage for that. Then raise next PR with tabs and coverage for that.

This might make sense as they are already separate tickets and generally we don't have PRs that cut across two tickets :)

server/controllers/temporary-accommodation/manage/v2/premisesController.ts
AvraamK: Fixed and yeah good point! I was initially testing redirection as I couldn't find an easy way to go to v2/ pages through the normal user-flow, but yeah true redirection is not needed!

server/services/v2/premisesService.ts
AvraamK: > as this is one is now only called by tests, perhaps this should just be

fair point, decided to go with that option in the end!

server/services/v2/premisesService.ts
AvraamK: added!

server/views/temporary-accommodation/v2/premises/index.njk
AvraamK: Fair point and yeah makes sense in terms of tracking of tests! Kept this PR focused only on what was asked and will do tabs coverage etc in the next one!

AvraamK:
> The designs in Miro show the search term in the online properties count if the use has searched for an address

thanks @robanderson-moj ! Updated PR with a fix for all 3 missing cases based on design:

1. zero items
2. has results - after search
3. zero results - after search

server/services/v2/premisesService.ts
gregkhawkins: if we are keeping this, maybe we should call it `searchDataAndGenerateTableRows()`

server/controllers/temporary-accommodation/manage/v2/premisesController.test.ts
```diff
@@ -64,43 +65,148 @@ describe('PremisesController', () => {
       { html: 'Manage' },
     ]
 
-    it('returns the table rows to the template', async () => {
-      const params: PremisesSearchParameters = { postcodeOrAddress: undefined }
+    it('defaults to status=online when no status parameter is provided', async () => {
```
gregkhawkins: perhaps this test should be called:

```
searches online premises data and returns as table rows when no status parameter is provided
```

server/controllers/temporary-accommodation/manage/v2/premisesController.test.ts
```diff
@@ -64,43 +65,148 @@ describe('PremisesController', () => {
...
+        isOnlineTab: true,
+        isArchivedTab: false,
+        ...searchData,
+      })
+
+      expect(premisesService.searchData).toHaveBeenCalledWith(callConfig, params, 'online')
+      expect(response.redirect).not.toHaveBeenCalled()
+    })
+
+    it('returns the search data to the template when status=online is provided', async () => {
```
gregkhawkins: then this can be called:
```
searches online premises data and returns as table rows when status=online parameter is provided
```

server/controllers/temporary-accommodation/manage/v2/premisesController.test.ts
```diff
@@ -64,43 +65,148 @@ describe('PremisesController', () => {
...
+        ...searchData,
       })
 
-      expect(premisesService.tableRows).toHaveBeenCalledWith(callConfig, params)
+      expect(premisesService.searchData).toHaveBeenCalledWith(callConfig, params, 'online')
     })
 
-    it('returns the filtered table rows to the template when the user has searched for a postcode or address', async () => {
-      const params = { postcodeOrAddress: 'NE1' }
+    it('returns the filtered search data to the template when the user has searched for a postcode or address', async () => {
```
gregkhawkins: then for consistency go with:
```
searches online premises data by postcode or address and returns as table rows when searched for a postcode or address
```

server/controllers/temporary-accommodation/manage/v2/premisesController.test.ts
```diff
@@ -64,43 +65,148 @@ describe('PremisesController', () => {
...
+        status: 'online',
+        isOnlineTab: true,
+        isArchivedTab: false,
+        ...searchData,
+      })
+
+      expect(premisesService.searchData).toHaveBeenCalledWith(callConfig, params, 'online')
+    })
+
+    it('returns empty search data when there are no properties in the database', async () => {
```
gregkhawkins, line 192: nice!

server/views/temporary-accommodation/v2/premises/index.njk
gregkhawkins: we still need integration tests for the counts as part of this and also are you moving the tabs to routes instead of param driven as per above comment? 
Forgive me if you are workiong on this in the background - I saw this was marked for a re-review so assumed this was the final PR. Do wonder whether as that refactor needs to happen we sh9ouod just exclude the tabs form this branch. Can have a chat maybe :)

gregkhawkins (changes requested):

server/views/temporary-accommodation/v2/premises/index.njk
```diff
@@ -35,43 +35,62 @@
...
+                </form>
+            </div>
         </div>
+        {% endif %}
+        {% if totalPremises is defined %}
+        <div class="govuk-grid-column-full">
+            <div class="govuk-!-margin-bottom-6">
+                <div class="govuk-!-padding-bottom-4">
+                    {% if params.postcodeOrAddress %}
+                    <h2 class="govuk-heading-l govuk-!-margin-bottom-6">{{ totalPremises }} {{ (isOnlineTab and "online" or "archived") }} properties matching '{{ params.postcodeOrAddress }}'</h2>
```
SaintRhys: Couldn't we just do `{{ status }}` here and below

server/views/temporary-accommodation/v2/premises/index.njk
```diff
@@ -98,10 +117,12 @@
...
         {% else %}
-            {% if params.postcodeOrAddress %}
-                <h2>{{ '0 online properties matching ‘' + params.postcodeOrAddress + '’' }}</h2>
-            {% else %}
-                <h2>0 online properties</h2>
+            {% if totalPremises is defined and totalPremises > 0 %}
+                {% if params.postcodeOrAddress %}
+                    <h2>{{ '0 online properties matching ‘' + params.postcodeOrAddress + '’' }}</h2>
+                {% else %}
+                    <h2>0 {{ (isOnlineTab and "online" or "archived") }} properties</h2>
```
SaintRhys, line 124: and here

server/controllers/temporary-accommodation/manage/v2/premisesController.ts
```diff
@@ -1,20 +1,29 @@
...
-      const tableRows = await this.premisesService.tableRows(callConfig, params)
-      return res.render('temporary-accommodation/v2/premises/index', { tableRows, params })
+      const searchData = await this.premisesService.searchDataAndGenerateTableRows(callConfig, params, status)
+
+      return res.render('temporary-accommodation/v2/premises/index', {
+        ...searchData,
+        params,
+        status,
+        isOnlineTab: status === 'online',
+        isArchivedTab: status === 'archived',
```
SaintRhys: I might have missed it but `isArchivedTab` isn't being used anywhere. Do we need to pass these extra params, could we just use status as needed.

server/views/temporary-accommodation/v2/premises/index.njk
```diff
@@ -35,43 +35,62 @@
...
-                }) }}
+    <div class="govuk-grid-row">
+        {% if (totalPremises is defined and totalPremises > 0) or params.postcodeOrAddress %}
+        <div class="govuk-grid-column-full">
+            <div class="moj-search cas-search-container govuk-!-margin-bottom-6 govuk-!-padding-6 govuk-!-padding-bottom-7 govuk-grid-column-two-thirds">
+                <form action="{{ paths.premises.v2[status]({}) }}" method="get">
+                    {{ govukInput({
+                        label: {
+                            classes: 'govuk-!-font-weight-bold',
+                            text: (isOnlineTab and 'Find an online property' or 'Find an archived property')
```
SaintRhys: Alternative way is to use status, like `text: 'Find an ' + status + ' property'`

SaintRhys (commented):
Looks good, just questioning the need for the isOnlineTab/isArchivedTab params.

server/paths/temporary-accommodation/manage.ts
```diff
@@ -4,6 +4,9 @@ const dashboardPath = temporaryAccommodationPath.path('dashboard')
 
 const premisesPath = temporaryAccommodationPath.path('properties')
 const premisesV2Path = temporaryAccommodationPath.path('v2/properties')
+const premisesV2OnlinePath = premisesV2Path.path('online')
+const premisesV2ArchivedPath = premisesV2Path.path('archived')
```
robanderson-moj, line 8: Are these paths going to cause issues when we add a page to view an individual premises?

I think we could get a clash with `v2/properties/{premisesId}` when we start working on [CAS-1737](https://dsdmoj.atlassian.net/browse/CAS-1737)

We might be better to keep the tab as a query parameter rather than a path

E.g.
`?status=online` and `?status=archived` rather than `/online` and `/archived`

robanderson-moj (changes requested):

server/controllers/temporary-accommodation/manage/v2/premisesController.ts
```diff
@@ -4,29 +4,25 @@ import { PremisesSearchParameters } from '@approved-premises/ui'
...
 
-      const params = req.query as PremisesSearchParameters & { status?: Cas3PremisesStatus }
-
-      let { status } = params
-      if (!status) {
-        status = 'online'
-      }
-
-      const searchData = await this.premisesService.searchData(callConfig, params, status)
+      const searchData = await this.premisesService.searchDataAndGenerateTableRows(callConfig, params, status)
```
gregkhawkins: can we just pass `params.postcodeOrAddress` instead of `params`

server/paths/temporary-accommodation/manage.ts
AvraamK: hmm @gregkhawkins what do you think on that? 
I can always revert them to query parameters as before. The whole logic was that this is how it's done in current bookings.
for CAS-1737 can't we have it like `v2/properties/id/{premisesId}` or similar?

server/paths/api.ts
```diff
@@ -92,6 +92,8 @@ export default {
   v2: {
     premises: {
       index: managePathsV2.premises.search,
+      online: managePathsV2.premises.search,
+      archived: managePathsV2.premises.search,
```
gregkhawkins: I maybe missing something but don't think we need these as these are paths for the api integrartion and all the same. Could rename `v2.premises.index` to `v2.premises.search` but also prob fine to be left as v2.premises.index`.

gregkhawkins (changes requested):

server/controllers/temporary-accommodation/manage/v2/premisesController.ts
robanderson-moj: Or add the status to the parameters?

gregkhawkins: status is now driven by the route. So, let's not build params up based on what the route gives us, otherwise might seem like status is a param passed on the url. 
Also, I would argue that services should have no concept of `params` / `headers` / `cookies` etc and the controllers interpret those and pass only the related data into the service level.

robanderson-moj: I worry that the status as a path parameter is going to cause issues when we add the view to show a single property - it might be better as a query parameter

https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/1322#discussion_r2171548518

server/paths/temporary-accommodation/manage.ts
robanderson-moj: Discussed in a huddle; happy with path parameters as it matches the way the rest of the application has been built

server/views/temporary-accommodation/v2/premises/index.njk
AvraamK: good point! done!

server/controllers/temporary-accommodation/manage/v2/premisesController.ts
AvraamK: Yeah status is simpler and better actually! done!

server/views/temporary-accommodation/v2/premises/index.njk
AvraamK: done!

server/paths/api.ts
AvraamK: done!

server/controllers/temporary-accommodation/manage/v2/premisesController.ts
gregkhawkins: as discussed with Rob and Avraam, sticking with routes instead of params

gregkhawkins (approved):

robanderson-moj (approved):

PR #1355, opened 2025-07-23
Archive bedspace - Form validation handling

AvraamK:
# Context

Implements the UI form validation handling for the CAS3 bedspace archiving feature. Users will have the ability to archive bedspaces with proper validation and error handling fully managed by the relevant backend service.
https://dsdmoj.atlassian.net/browse/CAS-1760

# Changes in this PR

- Added archive form and archive action routes to v2 bedspaces paths configuration
- Implemented archive method in v2 bedspace client to call POST `/cas3/premises/{premisesId}/bedspaces/{bedspaceId}/archive`
- Implemented archive controller methods with backend validation error handling
- Added routes for bedspace archive functionality
- Created archive form template with radio buttons for “Today” and “Another date” options
- Added archive action button to bedspace show page (visible for non-archived bedspaces only)
- Implemented specific error message handling for backend validation errors
- Added redirect to bedspace view page with green success banner on successful archive

### Before

Screenshot 2025-07-23 at 10 02 39

### After

Screenshot 2025-07-23 at 10 01 09

Screenshot 2025-07-23 at 10 01 26

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/integration_tests/tests/temporary-accommodation/manage/v2/bedspaceArchive.cy.ts
+++ b/integration_tests/tests/temporary-accommodation/manage/v2/bedspaceArchive.cy.ts
@@ -17,7 +17,7 @@ context('Bedspace Archive', () => {
     cy.task('stubRoomReferenceData')
   })
 
-  it('navigates to archive page, and archives successfully with success message based on the API response', () => {
+  it('Archives bedspace with today date option and shows success message', () => {
     // Given there is an active premises in the database
     const premises = cas3PremisesFactory.build({ status: 'online' })
     cy.task('stubSinglePremisesV2', premises)
@@ -49,7 +49,41 @@ context('Bedspace Archive', () => {
     finalBedspaceShowPage.shouldNotShowArchiveLink()
   })
 
-  it('navigates to archive page, and fails to archive with error based on the API response', () => {
+  it('uses the archive endpoint when user selects another date option and submits', () => {
+    // Given there is an active premises in the database
+    const premises = cas3PremisesFactory.build({ status: 'online' })
+    cy.task('stubSinglePremisesV2', premises)
+
+    // And there is an online bedspace in the database
+    const bedspace = cas3BedspaceFactory.build({ status: 'online' })
+    cy.task('stubBedspaceV2', { premisesId: premises.id, bedspace })
+
+    // When I visit the show bedspace page
+    const bedspaceShowPage = BedspaceShowPage.visit(premises, bedspace)
+
+    // Then I should see the archive button and be able to click it
+    bedspaceShowPage.clickArchiveLink()
+
+    // And I should navigate to the archive bedspace page
+    const archivePage = Page.verifyOnPage(BedspaceArchivePage, premises, bedspace)
+    archivePage.shouldShowBedspaceDetails()
+
+    // When I set up the API stub (success response)
+    cy.task('stubBedspaceArchiveV2', { premisesId: premises.id, bedspaceId: bedspace.id })
+    const archivedBedspace = { ...bedspace, status: 'archived' }
+    cy.task('stubBedspaceV2', { premisesId: premises.id, bedspace: archivedBedspace })
+
+    // And I select the another date option and enter a specific date
+    archivePage.selectAnotherDateOption()
+    archivePage.enterArchiveDate('15', '3', '2025')
+    archivePage.clickSubmit()
+
+    // Then the API should be called and I should be redirected with success
+    const finalBedspaceShowPage = Page.verifyOnPage(BedspaceShowPage, premises, archivedBedspace)
+    finalBedspaceShowPage.shouldShowBanner('Bedspace archived')
+  })
+
+  it('fails to archive with invalidEndDateInThePast error', () => {
     // Given there is an active premises in the database
     const premises = cas3PremisesFactory.build({ status: 'online' })
     cy.task('stubSinglePremisesV2', premises)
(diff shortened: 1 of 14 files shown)


server/i18n/en/errors.json
```diff
@@ -207,6 +207,15 @@
       "mustBeAtLeast1": "You must enter a duration of at least 1 day"
     }
   },
+  "bedspaceArchive": {
+    "endDate": {
+      "invalidEndDateInThePast": "The date is not within the next 3 months",
+      "invalidEndDateInTheFuture": "The date is not within the next 3 months",
```
robanderson-moj, line 215: Are both of these error messages meant to be the same? I couldn't see then in the Miro board or the ticket

server/data/v2/bedspaceClient.test.ts
```diff
@@ -74,4 +74,18 @@ describe('BedspaceClient', () => {
...
+    it('should archive a bedspace', async () => {
+      const bedspaceId = 'some-bedspace-id'
+      const archiveData = { endDate: '2024-12-31' }
+
+      fakeApprovedPremisesApi
+        .post(paths.cas3.premises.bedspaces.archive({ premisesId, bedspaceId }))
+        .matchHeader('authorization', `Bearer ${callConfig.token}`)
+        .reply(200)
+
+      await bedspaceClient.archive(premisesId, bedspaceId, archiveData)
```
robanderson-moj, line 88: Is there anything we can assert on here, or is the `archive` method not throwing an exception good enough?

server/i18n/en/errors.json
AvraamK: Initially it was supposed to be between last 7 days and next 3 months. But after some digging with the endpoint it was revealed that the past date, any past date (not just 7 days in the past), is just not supported. So the ticket description was agreed to be updated to only have that message.

server/data/v2/bedspaceClient.test.ts
AvraamK: sure can always add assertions!

server/controllers/temporary-accommodation/manage/v2/bedspacesController.ts
```diff
@@ -72,4 +80,76 @@ export default class BedspacesController {
...
+      const errors: Record<string, string> = {}
+
+      let archiveDate: string | undefined
+
+      if (archiveOption === 'today') {
+        archiveDate = DateFormats.dateObjToIsoDate(new Date())
+      } else if (archiveOption === 'anotherDate') {
+        if (dateIsBlank(req.body, 'archiveDate')) {
+          errors.archiveDate = 'You must specify the archive date'
+        } else if (!dateAndTimeInputsAreValidDates(req.body, 'archiveDate')) {
```
gregkhawkins: when I first saw this was thinking it might be overkill. This is because, when we send the date to the BE, Spring deserialises the date and in doing so will check the validity of it. It will send back a `400 Bad Request` when it fails which we could use to drive the `You must specify a valid archive date` messaging you have here.

I think Rhys added in this function recently when they asked us to add in some validation on the FE, this then lead to a discussions around validation in general. Off the back of that discussion we decided to not do any validation in the FE nd rely only on the BE because the usual benefits/justification for FE validation is that it does it in the client and so we get back quicker feedback. This is obviously not the case here as we do server-side rendering for the FE.

Anyway, this validation function was available I 100% understand why you made use of it and let's leave it as is - no need to take out - I just wanted to make sure you were aware that FE validation in general is a no no as the BE does it all and we can use the keys we get back to drive the messaging.

Think you already knew this! Just wanted to add in encase for clarity on ever evolving standards ;)

server/views/temporary-accommodation/v2/bedspaces/archive.njk
```diff
@@ -0,0 +1,80 @@
...
+  {{ showErrorSummary(errorSummary) }}
+  
+  <h1 class="govuk-heading-l">When should {{ bedspace.reference }} be archived?</h1>
+
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <form action="{{ paths.premises.v2.bedspaces.archive({ premisesId: params.premisesId, bedspaceId: bedspace.id })}}" method="post">
+        <input type="hidden" name="_csrf" value="{{ csrfToken }}"/>
+        
+        {% set conditionalHtml %}
```
gregkhawkins: minor: shall we call this `archiveDateHtml`?

integration_tests/tests/temporary-accommodation/manage/v2/bedspaceArchive.cy.ts
```diff
@@ -0,0 +1,254 @@
...
+
+    // And there is an online bedspace in the database
+    const bedspace = cas3BedspaceFactory.build({ status: 'online' })
+    cy.task('stubBedspaceV2', { premisesId: premises.id, bedspace })
+
+    // When I visit the show bedspace page
+    const bedspaceShowPage = BedspaceShowPage.visit(premises, bedspace)
+
+    // Then I should see the archive button and be able to click it
+    bedspaceShowPage.clickArchiveLink()
```
gregkhawkins: minor: to be clearer should this be `clickArchiveAction()`

server/views/temporary-accommodation/v2/bedspaces/archive.njk
```diff
@@ -0,0 +1,80 @@
...
+
+{% set pageTitle = "Archive a bedspace - " + applicationName %}
+{% set mainClasses = "app-container govuk-body" %}
+
+{% block beforeContent %}
+  {{ breadCrumb('Archive a bedspace', [
+    {title: 'List of properties', href: paths.premises.v2.online()},
+    {title: 'View a bedspace', href: paths.premises.v2.bedspaces.show({ premisesId: params.premisesId, bedspaceId: bedspace.id })}
+  ]) }}
+{% endblock %}
```
robanderson-moj: Based on the prototype and Miro, should this be a `Back` link to take the user back to the bedspace?

integration_tests/tests/temporary-accommodation/manage/v2/bedspaceArchive.cy.ts
```diff
@@ -0,0 +1,254 @@
...
+      bedspaceId: bedspace.id,
+      errorType: 'existingTurnaround',
+    })
+
+    // Archive with today option
+    archivePage.completeArchiveWithToday()
+
+    // Then I should see the validation error based on the API response
+    archivePage.shouldShowError('This bedspace cannot be archived due to a turnaround scheduled after this date.')
+  })
```
gregkhawkins: First off very thorough testing 👍 

_Couple of thoughts_...
1. We are covering all of the BE validation scenarios here but the extra ones you have added FE validation for (e.g.  blank date / invalid date) are not covered. When the FE valiidation fails we do a redirect to the `paths.premises.v2.bedspaces.archive` route where the passed error messages are rendered in the usual way.  Maybe we just need to cover one of these cases to check that all of this is wired up correctly so ensure the user gets the fail banner.
2. As I said above, very thorough testing, just wondering if too thorough? What I mean is, you have all of the BE validation cases covered here but I'm wondering whether one BE validation case is enough? I only wonder this as every test takes time in the pipeline (especially e2e/integration tests) so having the same test again and again with all validation keys we get from the BE might be overkill?

_Further musings_...
I might be wrong on (2) also, have you had a look at what we generally do for integration tests with lots of BE validation scenarios. If we always test them all (and take the hit in the pipeline) then perhaps this is the standard and ignore me. If you see lots of examples when we test only 1 BE validation scenario my suggestion might be worth considering.

server/controllers/temporary-accommodation/manage/v2/bedspacesController.ts
```diff
@@ -72,4 +80,76 @@ export default class BedspacesController {
...
+
+      if (!archiveDate) {
+        return res.redirect(paths.premises.v2.bedspaces.archive({ premisesId, bedspaceId }))
+      }
+
+      try {
+        const callConfig = extractCallConfig(req)
+        await this.bedspaceService.archiveBedspace(callConfig, premisesId, bedspaceId, archiveDate)
+
+        req.flash('success', 'Bedspace archived')
```
robanderson-moj, line 132: Does this message depend on when the entered date is?

The ticket says "Bedspace successfully archived", but the Miro has a number of different options

Image

gregkhawkins (commented):
Looking great overall - few comments.

Also, we need an e2e for the happy path for `Archive Bedspace`. This is possible now that @robanderson-moj has caught the e2e tests up for the rest of the `Manage properties V2` area

integration_tests/tests/temporary-accommodation/manage/v2/bedspaceArchive.cy.ts
SaintRhys: I'm more in favour of just a blanket test to just make sure we are showing whatever error the BE returns rather than individual errors. Can see something similar being used around, like in [premises](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/c646c7f367a046e7160c96da289597081f525473/integration_tests/tests/temporary-accommodation/manage/premises.cy.ts#L387). Or for individual fields like in [bedspace](https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/c646c7f367a046e7160c96da289597081f525473/integration_tests/tests/temporary-accommodation/manage/bedspace.cy.ts#L132).

gregkhawkins: looks like, based on @SaintRhys example he found, that my suggestion is probably in line with the standard so let's just have 1 integration test to cover the BE validation case (i.e. pick 1 message key) and 1 extra test to cover one of the the FE validation fail cases. So that this integration test gives us coverage on validation banners that are plumbed in the two different ways. Then unit tests should covers the individual scenarios which I think they already do!

server/controllers/temporary-accommodation/manage/v2/bedspacesController.ts
AvraamK: As discussed the detailed messages will be as part of a later ticket/branch

server/views/temporary-accommodation/v2/bedspaces/archive.njk
AvraamK: changed it! Used the previous one as it was used everywhere as well but it is indeed more clear now!

integration_tests/tests/temporary-accommodation/manage/v2/bedspaceArchive.cy.ts
AvraamK: done

server/controllers/temporary-accommodation/manage/v2/bedspacesController.ts
AvraamK: Hmm regarding this I was going to leave it as it is but I just discovered that the new FE integration test for the date was failing so in the end I removed completely the FE validation factor and from now we will rely on this one purely on BE. Apparently the response for empty or invalid date is enough to cover the empty date problem!

integration_tests/tests/temporary-accommodation/manage/v2/bedspaceArchive.cy.ts
AvraamK: Fair points all!
Updated the integration tests. Left only one BE one. No FE date-validation is required as this was just removed ( https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/1355#discussion_r2231535828 ) and now we rely 100% on BE for date validation as well!

AvraamK:
New PR: https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/1376

PR #1379, opened 2025-08-21
CAS 1870 Archive premises access rules

robanderson-moj:
# Context

Ticket [C3 FE: Handle Archive Premises Access Rules](https://dsdmoj.atlassian.net/browse/CAS-1870)
Requires `MANAGE_PROPERTIES_V2_ENABLED` feature flag

BE isn't currently removing duplicates in the response from `/cas3/premises/{premisesId}/can-archive`
BE is using the beds table for references rather than cas3 bedspaces table

# Changes in this PR

### After
image

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/server/controllers/temporary-accommodation/manage/v2/premisesController.ts
+++ b/server/controllers/temporary-accommodation/manage/v2/premisesController.ts
@@ -240,7 +240,18 @@ export default class PremisesController {
       const callConfig = extractCallConfig(req)
       const { premisesId } = req.params
 
-      const premises = await this.premisesService.getSinglePremises(callConfig, premisesId)
+      const [premises, blockingBedspaceReferences] = await Promise.all([
+        this.premisesService.getSinglePremises(callConfig, premisesId),
+        this.premisesService.canArchivePremises(callConfig, premisesId),
+      ])
+
+      if (blockingBedspaceReferences && blockingBedspaceReferences.affectedBedspaces.length > 0) {
+        return res.render('temporary-accommodation/v2/premises/cannot-archive', {
+          premises,
+          bedspaces: blockingBedspaceReferences.affectedBedspaces,
+        })
+      }
+
       const archiveOption = userInput.archiveOption || 'today'
 
       return res.render('temporary-accommodation/v2/premises/archive', {
--- a/cypress_shared/pages/temporary-accommodation/manage/v2/premisesCannotArchive.ts
+++ b/cypress_shared/pages/temporary-accommodation/manage/v2/premisesCannotArchive.ts
@@ -1,0 +1,22 @@
+import Page from '../../../page'
+import { Cas3BedspaceReference, Cas3Premises } from '../../../../../server/@types/shared'
+
+export default class PremisesCannotArchivePage extends Page {
+  constructor(premises: Cas3Premises) {
+    super(`You cannot archive ${premises.addressLine1}`)
+  }
+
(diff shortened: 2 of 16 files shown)


server/controllers/temporary-accommodation/manage/v2/premisesController.ts
```diff
@@ -240,7 +240,18 @@ export default class PremisesController {
       const callConfig = extractCallConfig(req)
       const { premisesId } = req.params
 
-      const premises = await this.premisesService.getSinglePremises(callConfig, premisesId)
+      const [premises, blockingBedspaceReferences] = await Promise.all([
+        this.premisesService.getSinglePremises(callConfig, premisesId),
+        this.premisesService.canArchivePremises(callConfig, premisesId),
+      ])
+
+      if (blockingBedspaceReferences && blockingBedspaceReferences.affectedBedspaces.length > 0) {
```
gregkhawkins, line 248: this is authorization for hitting this view in the first place really isn't it? Shame we don't have "guards" that we could add this to. Perhaps having this locking down the route in the long run might be worth considering.

Anyway, think this'll do for now especially with time constraints! As we redirect here (via the route) and then redirect again when "cannot archive" does the browser do this fast or does it look weird and not very performant?

gregkhawkins: we could put the `this.premisesService.canArchivePremises(callConfig, premisesId)` at the top so we do not need to wait for the other call before redirecting away?

I guess for the success case that makes this less performant as we are making the 2 calls synchronously. 
Anyway, if it's behaving itself let's stick with this for now :)

gregkhawkins (approved):

server/controllers/temporary-accommodation/manage/v2/premisesController.ts
robanderson-moj: This is less about authorization and more about stopping the user from wasting time on the archive page if there are bedspaces/bookings/voids that are going to prevent them from archiving the premises

From the user's perspective, we don't hit the response that renders the archive nunjucks page, it just renders cannot-archive nunjucks page instead.

robanderson-moj: The path in the user's browser doesn't change, it's just that `/archive` renders the `cannot-archive` view rather than letting them through to the archive form

gregkhawkins: okey doke. I think we are saying the same thing as "stopping the user from wasting time on the archive page if there are bedspaces/bookings/voids that are going to prevent them from archiving the premises" falls under the category of page authorization in my mind. I just mean this because we prevent them from accessing the page (as we do not want the to archive it).

All I mean really is that with most web frameworks you would have a more sophisticated way of preventing them from hitting the controller's `archive()` function in the first place by making this api call (e.g. called `guards` in Angular framework) so they never get the the controller in the first place. Anyway, nio need for us to build Angular framework but could consider an interceptor in the route as an alternative later down the line.

Anyway, sounds like it even if the browser route goes: `/archive` and then `/cannot-archive` this is not visible to the user from what you are saying 

Merge away!

robanderson-moj: Ah ok, I think I misunderstood what you were saying initially
We can make a note to have a look at adding a `guard` equivalent in the future, though would that turn this into two synchronous API calls rather than the current concurrent two?

PR #1382, opened 2025-08-22
Refactor manage properties v2 routes to v1

gregkhawkins:
The outcome of this spike: https://dsdmoj.atlassian.net/browse/CAS-1865 was to replace the current `Manage properties v1` routes with the new `Manage properties v2` routes.

**Included in this work is:**

* Refactor routes as described above
* Delete related v1 e2e tests (that are replaced by the Manage properties v2 work)
* Delete related v1 integration tests (that are replaced by the Manage properties v2 work)
* Point v2 e2e tests to new v1 routes
* Point v2 integration tests to new v1 routes
* Fix any e2e tests that relate to other areas that go into any of the new v2 areas (for setting up premises/bedspaces data etc)
* Fix any integration tests that relate to other areas that go into any of the new v2 areas (for setting up premises/bedspaces data etc)

**Not included in this work is:**
* Delete legacy v1 controllers/services related to this work
* Delete related v1 unit tests

The reason this is not included is we can do this once we go live and also these hold helpful example of functionality that is currently a WIP

PR #1390, opened 2025-08-28
Booking Confirmation and Arrivals area with Bedspace v2

gregkhawkins:
# Context

JIRA: https://dsdmoj.atlassian.net/browse/CAS-1928

The outcome of https://dsdmoj.atlassian.net/browse/CAS-1865 was to replace the current `Manage properties v1` routes with the new `Manage properties v2` routes. 

The above strategy helps us link into the other v1 areas of the application (e.g. `Bookings`, `Voids`, `Departures`, `Arrivals` etc) as the breadcrumbs still work (for the most part!) as they go back to v1 which is now the new v2 pages. 

However, these areas are currently driven by a `roomId` from the BE on the url path for the bedspace. With the way the new BE `cas3` endpoints have been delivered the primary key for bedspaces has become the `bedId` (not the `roomId`). This changes in the backend means that there is some refactoring to do in these `CAS3-UI` v1 areas to `retrieve the bedspace and not the room`. This has rippling effects on the v1 areas controllers, services and views and so we need to refactor these. 

# Changes in this PR
1. Get the `Booking confirmation` and `Arrivals` areas working
2. Add back in the related integration test / e2e test coverage that commented out in https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/1382

**Suggestion**: I have done separate commits for `Booking confirmation` area and `Arrivals` area so maybe easier to review 1 commit at a time

### DIFF
--- a/cypress_shared/pages/temporary-accommodation/manage/bookingArrivalEdit.ts
+++ b/cypress_shared/pages/temporary-accommodation/manage/bookingArrivalEdit.ts
@@ -1,4 +1,11 @@
-import type { Booking, LostBed, NewCas3Arrival as NewArrival, Premises, Room } from '@approved-premises/api'
+import type {
+  Booking,
+  Cas3Bedspace,
+  LostBed,
+  NewCas3Arrival as NewArrival,
+  Premises,
+  Room,
+} from '@approved-premises/api'
 import paths from '../../../../server/paths/temporary-accommodation/manage'
 import BedspaceConflictErrorComponent from '../../../components/bedspaceConflictError'
 import BookingInfoComponent from '../../../components/bookingInfo'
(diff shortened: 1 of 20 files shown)


cypress_shared/pages/temporary-accommodation/manage/bookingArrivalEdit.ts
```diff
@@ -15,22 +22,24 @@ export default class BookingArrivalEditPage extends Page {
...
+    this.bedspaceConflictErrorComponent = new BedspaceConflictErrorComponent(premises, room, bedspace, 'booking')
     this.bookingInfoComponent = new BookingInfoComponent(booking)
     this.popDetailsHeaderComponent = new PopDetailsHeaderComponent(booking.person)
     this.locationHeaderComponent = new LocationHeaderComponent({ premises, room })
   }
 
-  static visit(premises: Premises, room: Room, booking: Booking): BookingArrivalEditPage {
-    cy.visit(paths.bookings.arrivals.edit({ premisesId: premises.id, roomId: room.id, bookingId: booking.id }))
-    return new BookingArrivalEditPage(premises, room, booking)
+  static visit(premises: Premises, room: Room, bedspace: Cas3Bedspace, booking: Booking): BookingArrivalEditPage {
```
AvraamK, line 34: I guess it's fine for now and not PR-related, but I was wondering if it was worth to remove completely the whole 'room' factor if bedspace and room have essentially the same function? But this will be outside of scope of this ticket so it should be fine to stay as it is!

AvraamK (approved):

AvraamK:
all good here!

cypress_shared/pages/temporary-accommodation/manage/bookingArrivalEdit.ts
gregkhawkins: you're absolutely right. The problem is that the e2e test pulls out some information from the `View bedspace` page (after creating one) and then stores the `room` in memory of this type. Whereas the integration tests I refactored the data setup to create a `cas3Bedsapce` and pass that. As functions (e.g. this `visit()` one) are shared by both e2e and integration tests this enables the e2e test to pass the `room` and the integration test to pass the `bedspace`.

Once I finish all of the PRs to fix up all v1 sections, I will do a cleanup PR that gets the e2e test to pull out information from the `View bedspace` page and store in memory agasint a bedspace object of type `Cas3Bedspce` at that stage I am hoping I can rip out all references to `room` so that the application has no concept of `room` or `bed` and we are left with only `bedspace` 🤞 

Hope that all makes sense - and well spotted!

PR #1401, opened 2025-09-02
Manage properties V2 - Cleanup dead code for V1 from impl and unit tests

gregkhawkins:
We need to cleanup the leftovers from `Manage properties V1` which includes:
* any references to `room` as now moved to `bedspace`
* any references to `bed` as now moved to `bedspace`
* all premises v1 code (controlllers / services / utils)
* all bedspace v1 code (controlllers / services / utils)
* all references to `room` and `bed` in unit tests


AvraamK (approved):
Very clean! nothing to comment! all good!

PR #1447, opened 2025-10-07
CAS-2004: Add Gap Report button with feature flag

rashadmughal:


gregkhawkins (commented):
looking good, but probably want to add an e2e test (playwright) for this

gregkhawkins (approved):

PR #1457, opened 2025-10-16
CHORE: Remove excessive logs during integration tests

froddd:
This reduces the amount of logging produced during integration test runs, to make the output from the GitHub action and other loggers (own terminal, etc) more readable.

It also reduces the amount logged by default from the `debug` level to `info`. Should you wish to increase this default back to `debug`, the env var `LOG_LEVEL` can be set accordingly.


gregkhawkins (approved):
LGTM

PR #1456, opened 2025-10-16
Remove usages of v1 Premises and Bedspace services

froddd:
# Context

https://dsdmoj.atlassian.net/browse/CAS-2081 (Epic: https://dsdmoj.atlassian.net/browse/CAS-1562)

# Changes in this PR

This PR prepares the codebase before moving to the `/cas3/v2` Premises and Bedspace API endpoints. 

- Replaces all uses of deprecated `Premises` and `TemporaryAccommodationPremises` types (and associated factories) with the newer `Cas3Premises`
- Removes unused non-namespaced premises and bedspace API endpoints
- Removes any references to `v2` to provide a clean base for the move to the 'new' v2: services, clients and utilities have been combined where possible, and any leftover in `v2` directories have been moved back to the main component hierarchy
- Removes several unused methods wherever they were found (I think most of these were in integration test pages)
- `turnaroundWorkingDayCount` used for the `PUT` premises endpoint has been renamed to `turnaroundWorkingDays` to match all other requests -- I think @muhammad-elabdulla will be pleased about that 😉 

Would also like to go further with the cleanup and remove a lot of utilities that are found in services (such as https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/681239e28837224a34d8780bc8c09b43a9a3c97a/server/services/premisesService.ts#L165 and https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/blob/61ff818697cfb159bd24f82862b9517996669067/server/services/bedspaceService.ts#L195), as these really shouldn't appear there to keep concerns separated. However, this PR is becoming large enough as is, so I'll leave this for a later date!

## Screenshots of UI changes

No UI changes

# Release checklist

[Release process documentation](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/edit-v2/4247847062?draftShareId=a1c360ab-bd31-4db1-aae3-7cc002761de9).

## Pre-merge checklist

- [ ] Are any changes required to dependent services for this change to work?
  (eg. CAS API)
- [ ] Have they been released to production already?

## Post-merge checklist

- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to test
- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to preprod
[...]

PR #1461, opened 2025-10-21
Chore: remove last few v2 directories

froddd:
Follow-up to https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/1456 with a few more directories cleaned up.

Mostly straight file moves, but some combination going on with integration/e2e test pages in `cypress_shared/pages/temporary-accommodation/manage`, where the v2 test methods have been added to the original non-v2 pages. Those methods have been lifted verbatim with no functional changes at all, other than where needed (matching difering signatures, types, etc). The integration tests themselves have not had to be changed beyond imports.

Import/include/extend paths in views have been shortened to the absolute relative to the views root for the files moved, as this is considered a default views include and makes managing these paths a bit easier.


cypress_shared/pages/temporary-accommodation/manage/bedspaceEditable.ts
```diff
@@ -1,18 +1,18 @@
...
+  protected completeEditableForm(newOrUpdateBedspace: Cas3NewBedspace): void {
     this.getLabel('Enter a bedspace reference')
     this.getTextInputByIdAndEnterDetails('reference', newOrUpdateBedspace.reference)
 
-    if (this.bedStartDateIsEditable() && newOrUpdateBedspace.endDate) {
-      this.getLegend('Enter the bedspace end date (optional)')
-      this.completeDateInputs('startDate', newOrUpdateBedspace.endDate)
+    if (this.bedStartDateIsEditable() && newOrUpdateBedspace.startDate) {
+      this.getLegend('Enter the bedspace start date')
+      this.completeDateInputs('startDate', newOrUpdateBedspace.startDate)
```
froddd, line 11: I _think_ this is correct -- the original code didn't make a lot of sense, but do let me know if that's right!

server/views/temporary-accommodation/bedspaces/edit.njk
```diff
@@ -31,19 +31,20 @@
...
-                        },
+                        text: 'Bedspace reference',
+                        classes: "govuk-label--m"
+                    },
                         classes: "govuk-input--width-20",
                         hint: {
-                            text: "This will be used to identify the bedspace when making bookings and reporting on the service."
-                        },
+                        text: "This will be used to identify the bedspace when making bookings and reporting on the service."
+                    },
```
SaintRhys, line 46: Super small, might be an autoformatting thing but these indentations look off compared to the rest of the file/s, another in `_editable.njk` and there's a few more, with `html: dateHtml`

SaintRhys (commented):
Just a little one on formatting, feel free to ignore

server/views/temporary-accommodation/bedspaces/edit.njk
froddd: Ah yes, that is the auto formatting -- I believe the best way to avoid this is to ensure the helper opening is on one line, i.e.:

```nunjucks
{{ formPageInput({
  label: ''
  // ...
}) }}
```

instead of:
```nunjucks
{{ formPageInput(
  {
    label: '',
    // ...
  }
) }}
```

AvraamK: Awesome! This has been bugging me for some time now!

cypress.config.e2e.ts
```diff
@@ -51,6 +51,6 @@ export default defineConfig({
     acting_user_probation_region_id: 'ca979718-b15d-4318-9944-69aaff281cad',
     acting_user_probation_region_name: 'East of England',
   },
-  numTestsKeptInMemory: 0,
-  retries: 1,
+  numTestsKeptInMemory: process.env.CI ? 0 : 20,
```
AvraamK, line 54: nice!

AvraamK (approved):

cypress.config.e2e.ts
froddd: This means running E2E tests locally will be a bit faster when they fail, but it also enables clicking around on each step of the test to inspect values, DOM, etc. It was becoming absolutely essential working on these!

PR #1464, opened 2025-10-21
CAS-2081: use cas3 v2 endpoints

froddd:
# Context

https://dsdmoj.atlassian.net/browse/CAS-2081

# Changes in this PR

This PR enables the use of the new `/cas3/v2` API endpoints via the introduction of the `ENABLE_CAS3V2_API` feature flag.

The codebase was fully migrated to the newer types in order to be run wither with the flag off or on; this means that any entities of 'older' types coming from the non-v2 endpoints (and, in many other cases, even older endpoints) are now cast as the newer types, so everything will continue to work during the transition period.

The approach to this work has largely used the following process:

- Remove usages of older types and replace them with the newer types, as returned by the newer endpoints. This has introduced a lot of casting utilities, which will be removed in a subsequent PR when the migration has been fully deployed.
- Update clients to point to either the 'old' endpoints, or the newer `/cas3/v2` ones. This means the clients may return different types while the feature flag is actionable.
- Update services to cast older types as the newer ones. This includes any `POST` or `PUT` requests for creating/updating entities: the whole codebase now uses the newer request types, too.
- Update controllers and templates to use the newer types/entity structures.

Other than updating types and factories, the E2E and integration tests have been left largely untouched, to ensure the changes are not affecting functionality. Some exceptions to this:

- Reference data for premises and bedspace characteristics has been consolidated and is now based on the actual records available on preprod. This ensures the tests can use consistent data throughout the environments.
- a number of the mock APIs for integration tests now contain 2 versions of each endpoints: one for the `/cas3v2` endpoints, and another for the older ones.

I have tried to avoid making ad-hoc improvements to the implementation itself: some areas may seem like obvious candidates, but as this is a large refactor, it seemed a better idea to touch as little (!) of teh codebase as possible.

Where possible, utilities and conditionals thet enable the codebase to work with or without the feature flag enabled have been preceded with a `// TODO -- ENABLE_CAS3V2_API cleanup` comment -- this may help in determining what parts of the code should be removable once this work is deployed. However, I will also prepare a separate PR containing full removal of all conditionals.

## E2E tests

Both Cypress and PlayWright E2E tests are passing locally with the feature flag set to `false` or `true`.

## Screenshots of UI changes

This PR does not introduce any UI changes.

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod


integration_tests/mockApis/cancellation.ts
```diff
@@ -14,25 +16,52 @@ export default {
...
+  stubCancellationCreate: (args: {
+    premisesId: string
+    bookingId: string
+    cancellation: Cancellation
+  }): SuperAgentRequest =>
     stubFor({
       request: {
-        method: 'GET',
-        url: `/premises/${args.premisesId}/bookings/${args.bookingId}/cancellations/${args.cancellation.id}`,
+        method: 'POST',
```
froddd, line 59: The diff view is a bit out of whack here -- this should map to the old `stubCancellationCreate` above!

integration_tests/tests/temporary-accommodation/manage/bedspace.cy.ts
```diff
@@ -487,21 +492,11 @@ context('Bedspace', () => {
...
-        },
-        {
-          id: '7dd3bac5-3d1c-4acb-b110-b1614b2c95d8',
-          name: 'Shared kitchen',
-          serviceScope: 'temporary-accommodation',
-          modelScope: 'room',
-        },
+      const bedspaceCharacteristics: Array<Cas3BedspaceCharacteristic> = [
+        cas3BedspaceCharacteristicsFactory.byDescription('Ground floor level access'),
+        cas3BedspaceCharacteristicsFactory.byDescription('Shared kitchen'),
```
froddd, line 497: The factory for characteristics now returns more realistic characteristics based on the ones available on preprod at the moment: 'Single bed' didn't actually exist for bedspaces, so it's been swapped with another that does exist.

integration_tests/tests/temporary-accommodation/manage/booking.cy.ts
```diff
@@ -143,8 +142,12 @@ context('Booking', () => {
       expect(requests).to.have.length(1)
       const requestBody = JSON.parse(requests[0].body)
 
-      expect(requestBody.service).equal('temporary-accommodation')
-      expect(requestBody.bedId).equal(bedspace.id)
+      if (Cypress.env('ENABLE_CAS3V2_API') === true) {
```
froddd, line 145: Cypress doesn't like to access `config.flags` when running the integration tests, so instead we have to rely on this extra Cypress environment variable. The Github Actions workflow specifies this as `CYPRESS_ENABLE_CAS3V2_API`. The extra workflow will be fully removed when this is deployed.

server/controllers/temporary-accommodation/manage/bedspacesController.test.ts
```diff
@@ -45,8 +47,13 @@ describe('BedspacesController', () => {
...
   const lostBedId = 'some-lost-bed-id'
 
+  const displayedCharacteristics = [
+    cas3ReferenceDataFactory.build({ description: 'Self-catering' }),
+    cas3ReferenceDataFactory.build({ description: 'Suitable for foo' }),
+  ]
+
   const referenceData = {
-    characteristics: referenceDataFactory.characteristic('room').buildList(5),
+    characteristics: [...displayedCharacteristics, cas3ReferenceDataFactory.build({ description: 'Other' })],
```
froddd, line 56: There are some conditionals used when rendering the radio buttons that filter out the 'Other' option -- this wasn't explicitly tested, so this now covers it. I don't actually think the 'Other' option is shown or rendered anywhere, so it may be possible and/or desirable to set it to `active=false` in the DB?

server/controllers/temporary-accommodation/manage/bedspacesController.test.ts
```diff
@@ -233,39 +240,13 @@ describe('BedspacesController', () => {
...
-    }
 
     const onlineBedspace = cas3BedspaceFactory.build({
       id: bedspaceId,
       status: 'online',
       startDate: '2024-11-23',
-      characteristics: [characteristic1, characteristic2],
+      bedspaceCharacteristics: [characteristic1, characteristic2],
     })
-    const onlineBedspaceSummary: SummaryList = {
```
froddd, line 247: This output is tested in its own utility test suite, so we can simply assert it gets called with the right arguments.

server/controllers/temporary-accommodation/manage/bookingsController.ts
```diff
@@ -207,19 +208,16 @@ export default class BookingsController {
       const { arrivalDate } = DateFormats.dateAndTimeInputsToIsoString(req.body, 'arrivalDate')
       const { departureDate } = DateFormats.dateAndTimeInputsToIsoString(req.body, 'departureDate')
 
-      const bedspace = await this.bedspacesService.getSingleBedspace(callConfig, premisesId, bedspaceId)
```
froddd, line 210: The bedspace wasn't actually used for anything other than its `id`, which we already have as `bedspaceId` from the path params, so we save a call to the API here.

server/controllers/temporary-accommodation/manage/premisesController.test.ts
```diff
@@ -75,42 +86,25 @@ describe('PremisesController', () => {
     })
     ;(extractCallConfig as jest.MockedFn<typeof extractCallConfig>).mockReturnValue(callConfig)
     assessmentService.findAssessment.mockResolvedValue(assessment)
+
+    jest.spyOn(premisesUtils, 'tableRows')
+    jest.spyOn(premisesUtils, 'summaryList')
   })
 
   describe('index', () => {
-    const premisesRow1 = [
```
froddd, line 81: `tableRows` is covered in its own test suite, so no need for this fixture data here.

server/controllers/temporary-accommodation/manage/premisesController.test.ts
```diff
@@ -120,7 +114,7 @@ describe('PremisesController', () => {
         path: '/properties/online',
       })
 
-      premisesService.searchDataAndGenerateTableRows.mockResolvedValue(searchData)
+      premisesService.search.mockResolvedValue(searchData)
```
froddd, line 117: The clients have largely been stripped of any formatting/transforming duties, to make it easier to switch between testing with flag on and off -- these transforms have been moved either to the respective services, or to separate utilities.

server/data/bedspaceClient.test.ts
```diff
@@ -196,4 +205,38 @@ describeClient('BedspaceClient', provider => {
       expect(result).toEqual(bedspace)
     })
   })
+
+  describe('with the ENABLE_CAS3V2_API flag off', () => {
```
froddd: The tests contained in the `with the ENABLE_CAS3V2_API flag off` describe blocks are as close to what the original tests were as possible, to ensure no functionality changes. Unfortunately the diff will show them all as new additions... I have tried to move things around to produce a better diff, but no luck. Soz 🤷‍♂️

server/data/bedspaceClient.ts
```diff
@@ -18,8 +19,15 @@ export default class BedspaceClient {
...
 
   async search(searchParameters: Cas3BedspaceSearchParameters) {
-    return this.restClient.post<Cas3BedspaceSearchResults>({
-      path: paths.bedspaces.search({}),
+    if (!config.flags.enableCas3v2Api) {
+      return this.restClient.post<Cas3BedspaceSearchResults>({
+        path: paths.cas3.bedspaces.search({}),
+        data: searchParameters,
+      })
+    }
```
froddd, line 27: Using a negative conditional here should make removal of the old branches a lot easier. The new world is the default now!

server/paths/api.ts
froddd, line 1: Quite a bit of a refactor here: it was getting difficult to parse which paths were CAS3-namespaced and which ones were not, given a lot of the transitory objects setup for `assessPaths`, `applicationPaths`, etc. I decided to flatten things a fair bit and it did help during development -- with the unfortunate side effect that it makes this diff difficult to parse! I'm hoping the recent update to the client tests, bringing in proper contract testing, will remove any nervousness about the validity of all the paths defined here 🙂

integration_tests/mockApis/extension.ts
```diff
@@ -4,8 +4,63 @@ import type { Extension } from '@approved-premises/api'
 
 import { getMatchingRequests, stubFor } from '.'
 import { bedspaceConflictResponseBody, errorStub } from './utils'
+import paths from '../../server/paths/api'
+import config from '../../server/config'
 
-export default {
+const cas3V2ApiEnabledStubs = {
```
SaintRhys: Minor but should this be `cas3v2ApiEnabledStubs`

server/testutils/factories/cas3Booking.ts
```diff
@@ -0,0 +1,135 @@
...
+}
+
+export default Cas3BookingFactory.define(() => {
+  const originalArrivalDate = faker.date.soon()
+  const arrivalDate = faker.date.soon()
+  const departureDate = faker.date.soon({ days: 84, refDate: arrivalDate })
+  const turnaroundStartDate = faker.date.soon({ days: 1, refDate: departureDate })
+  const effectiveEndDate = faker.date.soon({ days: 5, refDate: departureDate })
+
+  const cancellations = faker.helpers.arrayElements(cancellationFactory.buildList(5))
```
SaintRhys: Could these be using cas3 factories

e2e/tests/stepDefinitions/manage/booking.ts
```diff
@@ -31,7 +31,7 @@ Given('I create a booking with all necessary details', () => {
         crn: person.crn,
       })
 
-      const booking = bookingFactory.provisional().build({
+      const booking = cas3BookingFactory.provisional().build({
```
SaintRhys, line 34: Could the below be cas3 factory for turnaround

server/testutils/factories/cas3Premises.ts
```diff
@@ -65,7 +67,8 @@ export default Cas3PremisesFactory.define(() => ({
   totalOnlineBedspaces: faker.number.int({ min: 1, max: 10 }),
   totalArchivedBedspaces: faker.number.int({ min: 1, max: 5 }),
   totalUpcomingBedspaces: faker.number.int({ min: 1, max: 5 }),
-  characteristics: characteristicFactory.buildList(5),
+  characteristics: undefined as Array<Characteristic>,
```
SaintRhys: Would this be better as an empty array or is it needed as undefined

SaintRhys (commented):
Looking good, some solid changes.

There's a `Booking` in `bookingSearch` (cypress_shared) that could be swapped to `Cas3Booking`

and a couple bits about old factories in new ones.

server/testutils/factories/cas3Booking.ts
froddd: Yes, good spot! They absolutely should, in fact. I'm surprised the typecheck isn't actually picking this up 🤔

server/testutils/factories/cas3Premises.ts
froddd: The new API doesn't return this property at all, so `undefined` is correct (it populates the `premisesCharacteristics` instead). But you make a good point -- it should probably not be set at all, so I'm removing it fully, and also removing the equivalend in the bedspace factory.

e2e/tests/stepDefinitions/manage/booking.ts
froddd: Absolutely, fixed!

integration_tests/mockApis/extension.ts
froddd: Eagle eyes! Thanks 🙂

server/utils/bedspaceUtils.test.ts
```diff
@@ -142,4 +152,308 @@ describe('bedspaceV2Utils', () => {
       })
     })
   })
+
+  describe('summaryList', () => {
```
froddd, line 156: The Summary List-related tests below have been moved from the test suite for the Bedspace Service, where the implementation originally resided.

server/utils/bedspaceUtils.ts
```diff
@@ -133,3 +141,141 @@ export function setDefaultStartDate(userInput: Record<string, unknown>) {
     userInput['startDate-year'] = String(today.getFullYear())
   }
 }
+
+export const summaryList = (bedspace: Cas3Bedspace): SummaryList => {
```
froddd, line 145: These methods have been moved from the Bespace Service.

server/utils/premisesUtils.test.ts
```diff
@@ -218,4 +234,408 @@ describe('premisesUtils', () => {
       expect(isPremiseScheduledToBeArchived(totals)).toBe(true)
     })
   })
+
+  describe('tableRows', () => {
```
froddd, line 238: Tests and implementations moved from the Premises Service.

SaintRhys (dismissed):

AvraamK (dismissed):

SaintRhys (approved):

PR #1458, opened 2025-10-17
Show error message when users account is missing staff details

AvraamK:
# Context

Jira: https://dsdmoj.atlassian.net/browse/CAS-1311

# Changes in this PR

Added a new screen that notify users if there is an issue with their account

## Screenshots of UI changes
Screenshot 2025-10-17 at 14 00 18

# Release checklist

[Release process documentation](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/edit-v2/4247847062?draftShareId=a1c360ab-bd31-4db1-aae3-7cc002761de9).

## Pre-merge checklist

- [ ] Are any changes required to dependent services for this change to work?
  (eg. CAS API)
- [ ] Have they been released to production already?

## Post-merge checklist

- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to test
- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to preprod
- [ ] [Manually approve](https://dsdmoj.atlassian.net/wiki/spaces/AP/pages/4247847062/Release+process#Manual-releases)
  release to prod

PR #1467, opened 2025-10-22
Add error message on users without CAS3 access

AvraamK:
# Context

Jira comment: 
https://dsdmoj.atlassian.net/browse/CAS-1311?focusedCommentId=673308

# Changes in this PR

Swapped the access logic for nDelius and CAS3.
nDelius users without a CAS3 role now see the new "Update your account" page that was first introduced here (merged: https://github.com/ministryofjustice/hmpps-temporary-accommodation-ui/pull/1458 ).
Every other user see the "non authorised" page

Before it was the opposite.

## How to test:

Case 1: Login as a user with nDelius access but without CAS3 access: should see the new "Update your account" page.

Case 2: Login as a user without nDelius or CAS3 access: should see the existing "not authorised" page

## Screenshots of UI changes

updateyouraccount

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/cypress_shared/pages/userDetailsRequired.ts
+++ b/cypress_shared/pages/userDetailsRequired.ts
@@ -1,0 +1,7 @@
+import Page from './page'
+
+export default class UserDetailsRequiredPage extends Page {
+  constructor() {
+    super('Update your account details to use this service')
+  }
+}
--- a/integration_tests/tests/landing.cy.ts
+++ b/integration_tests/tests/landing.cy.ts
@@ -1,5 +1,5 @@
 import { ListPage } from '../../cypress_shared/pages/apply'
-import NotAuthorisedPage from '../../cypress_shared/pages/notAuthorised'
+import UserDetailsRequiredPage from '../../cypress_shared/pages/userDetailsRequired'
 import Page from '../../cypress_shared/pages/page'
 import DashboardPage from '../../cypress_shared/pages/temporary-accommodation/dashboardPage'
 import { setupTestUser, setupTestUserWithoutRole } from '../../cypress_shared/utils/setupTestUser'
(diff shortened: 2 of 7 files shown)


froddd (commented):
Looks great code-wise, but the wording on the new page seems wrong for this use case: if we're showing this page to any user who has neither `assessor` nor `referrer` role, should we tell them to ask someone to add the `assessor` role? What if they only need `referrer`?

AvraamK:
> Looks great code-wise, but the wording on the new page seems wrong for this use case: if we're showing this page to any user who has neither `assessor` nor `referrer` role, should we tell them to ask someone to add the `assessor` role? What if they only need `referrer`?

yeah I was thinking about that. But at the same time the most usual cases will be these 2 roles right?
If not then we might need a more neutral message to include all roles(?). Page is supposed to be as static as possible as we don't get more info from BE on what we actually missing, since it is an auth issue.

froddd:
Yeah, maybe a more generic message would work better -- 'ask your line manager or HPT lead to add the relevant role to your account'?

AvraamK:
> Yeah, maybe a more generic message would work better -- 'ask your line manager or HPT lead to add the relevant role to your account'?

hey @froddd  I've just confirmed that PPs get the Referrer role by default so the text we are showing is the correct one!

froddd:
@AvraamK the integration tests say "redirects to the `User Details Required` page if the user is neither a assessor or a referrer". 

If a referrer user has neither the 'referrer' nor the 'assessor' role, they will see a page which tells them to ask their line manager or HPT lead to add the 'assessor' role to their account. This is incorrect: the line manager or HPT lead should only add the 'referrer' role to their account in this case, which is why I suggested the more generic wording.

Either the tests (and implementation?) are incorrect, or the wording is incorrect? In any case, those things don't line up!

AvraamK:
> @AvraamK the integration tests say "redirects to the `User Details Required` page if the user is neither a assessor or a referrer".
> 
> If a referrer user has neither the 'referrer' nor the 'assessor' role, they will see a page which tells them to ask their line manager or HPT lead to add the 'assessor' role to their account. This is incorrect: the line manager or HPT lead should only add the 'referrer' role to their account in this case, which is why I suggested the more generic wording.
> 
> Either the tests (and implementation?) are incorrect, or the wording is incorrect? In any case, those things don't line up!

oh right now I get it! I thought originally we would had only one scenario!
Yeah in that way the neutral wording is the correct way.
Just updated it!

thanks again for spotting this!

froddd (approved):
👍

PR #1469, opened 2025-10-27
Add Pact tests for clients

SaintRhys:
# Context

[This](https://dsdmoj.atlassian.net/browse/CAS-2088) adds api contract testing via pacts, to bring it in line with other cas.

This also includes a cleanup of unneeded code.

- removed `bookings`, in reportClient
- remove `pipeBookings` in reportService
- removed `document` in personClient
- remove `getDocument` in personService
- removed `risks` in personClient
- remove `getPersonRisks` from personService
- remove `findCancellation` from bookingClient
- remove `findDeparture` from bookingClient
- removed `getDeparture` from departureService
- removed `getDocument` from applicationService
- removed `getCancellation` from cancellationService
- removed `documents` from applicationClient

# Changes in this PR

### After

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod


server/data/assessmentClient.test.ts
```diff
@@ -46,59 +41,90 @@ describe('AssessmentClient', () => {
...
-
         expect(result.data).toEqual(assessments.data)
-        expect(result.pageNumber).toEqual(assessments.pageNumber)
-        expect(result.pageSize).toEqual(assessments.pageSize)
-        expect(result.totalPages).toEqual(assessments.totalPages)
-        expect(result.totalResults).toEqual(assessments.data.length)
+        // expect(result.pageNumber).toEqual(assessments.pageNumber)
+        // expect(result.pageSize).toEqual(assessments.pageSize)
+        // expect(result.totalPages).toEqual(assessments.totalPages)
+        // expect(result.totalResults).toEqual(assessments.data.length)
```
froddd: Can be removed!

server/data/applicationClient.ts
```diff
@@ -23,11 +23,13 @@ export default class ApplicationClient {
     })
   }
 
-  async create(crn: string, activeOffence: ActiveOffence) {
+  async create(crn: string, activeOffence: ActiveOffence, createWithRisks?: boolean) {
```
froddd: Other than tests, there's no instance of this being called with `createWithRisks`, could it be removed altogether? I think the previous implementation relying solely on `config.flags.oasysDisabled` should be sufficient?

If this was added only to allow testing, could the test be rewritten with a `config.flags` override?

froddd (dismissed):
Very nice! The cleanup is most excellent, and this really shows how far some of the factories had diverged from reality. This should bring heaps of confidence that the UI is absolutely in line with the API. 

Only one question around the reimplementation of handling of the `createWithRisks` query param, which could possibly be avoided by tweaking the test instead?

server/data/assessmentClient.test.ts
SaintRhys: Missed that!

server/data/applicationClient.ts
SaintRhys: Good shout, reworked  it. I'm wondering if we need `createWithRisks` at all, can't see us setting the flag.

froddd: It's a good point -- I believe it's a leftover from early development on CAS1, and it was actually removed from CAS1 not long ago -- let's create a separate PR to remove it, to make it easy to undo -- just in case!

froddd (approved):
Looks good!

PR #1471, opened 2025-10-28
CAS-2005: C3 FE Gap report date range error

rashadmughal:
Show error for gap report when date range is too >31 days

### DIFF
--- a/server/i18n/en/errors.json
+++ b/server/i18n/en/errors.json
@@ -102,6 +102,7 @@
       "beforeStartDate": "The end date must be before the start date",
       "inFuture": "The end date must be today or a past date",
       "rangeTooLarge": "The end date cannot be more than 3 months after the start date",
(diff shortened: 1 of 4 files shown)


server/i18n/en/errors.json
```diff
@@ -102,6 +102,7 @@
       "beforeStartDate": "The end date must be before the start date",
       "inFuture": "The end date must be today or a past date",
       "rangeTooLarge": "The end date cannot be more than 3 months after the start date",
+      "rangeTooLargeGapReport": "The end date cannot be more than 31 days from the start date",
```
gregkhawkins, line 105: at the risk of being a pain, is it worth an e2e for that use case?

robanderson-moj:
Can you rebase this branch before I review?

server/i18n/en/errors.json
rashadmughal: yes as discussed I will add an integration test

robanderson-moj (approved):
👍

PR #1475, opened 2025-10-30
Chore: Fix issue with samesite cookies inside Cypress UI

froddd:
A [recent change to the Auth service](https://github.com/ministryofjustice/hmpps-auth/commit/6829de75efc17fc453282e1cb4e6c6751d63a3df) specifies `samesite=Strict` for auth cookies. This breaks flows in Cypress UI, as those are run inside a frame for the given URL, which means they never pass the samesite validation. This change intercepts requests containing those auth cookies to remove the restriction.

This commit also removes a wait and navigateto the dashboard, which is seemingly no longer needed. This may save milliseconds to our test runs 🙂

### DIFF
--- a/e2e/tests/stepDefinitions/manage.ts
+++ b/e2e/tests/stepDefinitions/manage.ts
@@ -24,15 +24,22 @@ const signIn = (usernameVariable: string, passwordVariable: string) => {
   const username = Cypress.env(usernameVariable) || throwMissingCypressEnvError(usernameVariable)
   const password = Cypress.env(passwordVariable) || throwMissingCypressEnvError(passwordVariable)
 
+  // If the test is run in Cypress UI mode, rewrite auth cookies with strict or lax samesite policy
+  // to ensure they work from the Cypress iframe.
+  // https://www.tomoliver.net/posts/cypress-samesite-problem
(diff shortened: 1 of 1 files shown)


e2e/tests/stepDefinitions/manage.ts
```diff
@@ -24,15 +24,22 @@ const signIn = (usernameVariable: string, passwordVariable: string) => {
   const username = Cypress.env(usernameVariable) || throwMissingCypressEnvError(usernameVariable)
   const password = Cypress.env(passwordVariable) || throwMissingCypressEnvError(passwordVariable)
 
+  // If the test is run in Cypress UI mode, rewrite auth cookies with strict or lax samesite policy
+  // to ensure they work from the Cypress iframe.
+  // https://www.tomoliver.net/posts/cypress-samesite-problem
+  if (!Cypress.env('CYPRESS_RUN_HEADLESS')) {
```
robanderson-moj: We're going to need to change this check, I can't currently run `npm run test:e2e` locally.

Maybe we could use the `environment === 'local'` check you removed further down?

robanderson-moj (approved):
:+1:

PR #1487, opened 2025-11-10
Chore: Ensure correct Local Authority is selected for E2E premises test

froddd:
The Local Authority selected when creating or editing a premises is based on a random pick from the list of available options in the local authority dropdown select element.

If this pick happened to be 'York', the test would type 'York', then click on the first autocomplete option shown: in this case, this would be 'East Riding of Yorkshire'. This would cause a discrepancy between the value used to assert the correct details have been entered and the actual value entered.

To fix this, we ensure the autocomplete option clicked on is an exact match for the value randomly selected, rather than the first one shown.

### Autocomplete filled with 'York', showing first option is different:

Screenshot 2025-11-10 at 11 44 28


robanderson-moj (approved):
👍

PR #1491, opened 2025-11-12
chore: CAS-2081 -- cleanup enable cas3v2 api flag

froddd:
This PR completely removes implementations linked to the `ENABLE_CAS3V2_API` flag being set to `false`. It also removes any unused methods and factories as a result of the drop of implementation, and removes any unused non-namespaced API paths currently defined.

No UI changes.


server/data/bookingClient.test.ts
froddd, line 1: This file is effectively the deleted file above renamed to the old test suite -- Git/GitHub didn't quite get that!

server/data/lostBedClient.test.ts
froddd, line 1: Same as for `bookingClient.test.ts`, this is effectively a rename of the preceding file.

SaintRhys (approved):
Big cleanup, looks good

PR #1494, opened 2025-11-13
CAS-2091 Add maintenance mode

robanderson-moj:
# Context

Ticket https://dsdmoj.atlassian.net/browse/CAS-2091

Set the environment variable `IN_MAINTENANCE_MODE` to `true` to enable for all users who don't have the `CAS3_ADMIN` role: https://github.com/ministryofjustice/hmpps-approved-premises-api/pull/4433

# Changes in this PR

### After
image

# Release checklist

## Pre-merge checklist

  (eg. CAS API)

## Post-merge checklist

  release to test
  release to preprod
  release to prod

### DIFF
--- a/server/utils/userUtils.ts
+++ b/server/utils/userUtils.ts
@@ -23,6 +23,10 @@ export function userHasReporterRole(user: User): boolean {
   return user.roles.includes('reporter')
 }
 
+export function userHasAdminRole(user: User): boolean {
+  return user.roles.includes('admin')
+}
+
 export function userHasReferrerRoleAndIsApplyEnabled(user: User): boolean {
   return userHasReferrerRole(user) && isApplyEnabledForUser(user)
 }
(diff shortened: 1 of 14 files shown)


integration_tests/tests/maintenance.cy.ts
froddd, line 1: Given the difficulty in testing different environment configurations in integration tests, we should be ok leaving integration tests out altogether. The middleware test suite should be enough, and given both CAS1 and CAS2 do the same, might as well keep things consistent between repositories!

server/utils/userUtils.ts
```diff
@@ -12,15 +12,19 @@ export function isApplyEnabledForUser(user: User): boolean {
 }
 
 export function userHasReferrerRole(user: User): boolean {
-  return user.roles.includes('referrer')
+  return user.roles.includes('referrer') || userHasAdminRole(user)
```
froddd: Given a user can have multiple roles, are these changes actually necessary? I can see that they effectively give all roles to anyone with the role 'admin', but we may want to keep this control granular? At the very least this should be covered by tests, so the logic is documented.

robanderson-moj: Noted, and updated

froddd (approved):
Looks good!

PR #1511, opened 2025-12-01
Chore: Install and configure hmpps npm-script-allowlist

froddd:
https://github.com/ministryofjustice/hmpps-typescript-lib/tree/main/packages/npm-script-allowlist

This configures some safer defaults for NPM, ensuring only explicitly allowed packages can run postinstall scripts.


robanderson-moj (approved):
👍

PR #1512, opened 2025-12-01
Ensure unit test future date is in future

froddd:
A unit test was setting a future date of `2025-12-01`, which happens to be today. As a result, the test started failing.

This PR ensures the date used in the unit test is always a future date.


robanderson-moj (approved):
Looks good

PR #1524, opened 2025-12-12
HAAR-3945: 🔧 Move hmpps-auth to ghcr

simon-mitchell:


froddd (approved):
Yep, thanks! Good to merge

PR #1525, opened 2025-12-15
Chore: update node

froddd:
This updates `node` to v24 and `npm` to v11. The `node-setup` github action has also been updated to v6 all around.


bobmeredith (approved):
Seems to work fine!
Thanks for this.

PR #1530, opened 2025-12-17
Security fix - stop logging username (log the user's ID instead)

gregkhawkins:
# Context

* JIRA: https://dsdmoj.atlassian.net/browse/CAS-1653
* Discovered that we were logging usernames (which in most cases are personal emails)

# Changes in this PR
* Log `user.id` in the error logs where we were previously logging `user.username`

### DIFF
--- a/server/middleware/populateCurrentUser.ts
+++ b/server/middleware/populateCurrentUser.ts
@@ -28,7 +28,7 @@ export default function populateCurrentUser(userService: UserService): RequestHa
         logger.error('Delius account missing staff details')
         return res.redirect('/deliusMissingStaffDetails')
       }
-      logger.error(error, `Failed to retrieve user for: ${res.locals.user && res.locals.user.username}`)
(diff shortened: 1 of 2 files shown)


server/middleware/populateCurrentUser.ts
```diff
@@ -28,7 +28,7 @@ export default function populateCurrentUser(userService: UserService): RequestHa
         logger.error('Delius account missing staff details')
         return res.redirect('/deliusMissingStaffDetails')
       }
-      logger.error(error, `Failed to retrieve user for: ${res.locals.user && res.locals.user.username}`)
+      logger.error(error, `Failed to retrieve user for: ${res.locals.user && res.locals.user.id}`)
```
froddd, line 31: Could probably shorthand this a bit:


```suggestion
      logger.error(error, `Failed to retrieve user for: ${res.locals.user?.id}`)
```

froddd (approved):
Nice!

PR #1544, opened 2026-01-19
SAS-220: replace lost bed ref data endpoint

froddd:
# Context

https://dsdmoj.atlassian.net/browse/SAS-220

# Changes in this PR

This PR changes the endpoint used to fetch the Void Bedspace reasons reference data from `/reference-data/lost-bed-reasons` to `/cas3/reference-data?type=VOID_BEDSPACE_REASONS`.

In addition, this creates a new definition for the existing `/reference-data` endpoint, and updates hardcoded references, to make usages more discoverable.

A test has been updated to assert that the reasons are returned in the correct order.

## Screenshots of UI changes

No UI changes.


robanderson-moj (approved):
👍 Looks good

PR #1548, opened 2026-01-21
Show no overstay if there is a more recent extension

robanderson-moj:
# Context

Show no overstay if the booking departure date has been readjusted back to within 84 days

# Changes in this PR

### After
image
image
image

### DIFF
--- a/server/utils/bookingUtils.ts
+++ b/server/utils/bookingUtils.ts
@@ -410,18 +410,32 @@ const assessmentRadioItemText = (assessmentSummary: Cas3AssessmentSummary) => {
 }
 
 export const getOverstaySummary = (booking: Cas3Booking): string => {
-  const { arrivalDate, overstays } = booking
+  const { arrivalDate, overstays, extensions } = booking
 
   if (overstays.length === 0) {
     return 'No'
   }
 
-  const sortedOverstays = overstays.sort((a, b) => a.createdAt.localeCompare(b.createdAt)).reverse()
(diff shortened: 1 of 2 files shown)


server/utils/bookingUtils.ts
```diff
@@ -410,18 +410,32 @@ const assessmentRadioItemText = (assessmentSummary: Cas3AssessmentSummary) => {
...
 
-  const sortedOverstays = overstays.sort((a, b) => a.createdAt.localeCompare(b.createdAt)).reverse()
+  const sortedOverstays = overstays.sort(sortByCreatedAtTimestamp).reverse()
+  const latestOverstay = sortedOverstays[0]
+
+  if (extensions.length > 0) {
+    const sortedExtensions = extensions.sort(sortByCreatedAtTimestamp).reverse()
+    const latestExtension = sortedExtensions[0]
+
+    if (sortByCreatedAtTimestamp(latestExtension, latestOverstay) > 0) {
```
froddd: Wonder if it might be a tad clearer if this was called `compareCreatedAt` instead of `sortByCreatedAtTimestamp`?

froddd (dismissed):
Nice catch on that functionality! One small suggestion for readability, no biggie.

server/utils/bookingUtils.ts
robanderson-moj: Good call, I've changed the name now

froddd (approved):

PR #1555, opened 2026-01-30
Booking overstay departures

robanderson-moj:
# Context

This adds to the overstay functionality by allowing users to mark a booking as both overstayed and departed at the same time.

Based on https://dsdmoj.atlassian.net/jira/software/c/projects/CAS/boards/11445?selectedIssue=CAS-2140

This adds the `Mark booking as departed` route from the overstayers Miro board

# Changes in this PR

### Before
image

### DIFF
--- a/integration_tests/tests/temporary-accommodation/manage/overstay.cy.ts
+++ b/integration_tests/tests/temporary-accommodation/manage/overstay.cy.ts
@@ -7,37 +7,20 @@ import { setupBookingStateStubs } from '../../../../cypress_shared/utils/booking
 import { setupTestUser } from '../../../../cypress_shared/utils/setupTestUser'
 import {
   cas3BookingFactory,
+  cas3DepartureFactory,
+  cas3NewDepartureFactory,
   cas3NewOverstayFactory,
   cas3OverstayFactory,
   newExtensionFactory,
 } from '../../../../server/testutils/factories'
 import { DateFormats } from '../../../../server/utils/dateUtils'
 import Cas3NewOverstay from '../../../../server/testutils/factories/cas3NewOverstay'
+import BookingDepartureNewPage from '../../../../cypress_shared/pages/temporary-accommodation/manage/bookingDepartureNew'
 
 context('Booking overstay', () => {
   beforeEach(() => {
     cy.task('reset')
     setupTestUser('assessor')
-  })
-
-  it('navigates to the overstay page via the booking extension page', () => {
-    cy.signIn()
-
-    const booking = cas3BookingFactory.arrived().build()
-    const { premises, bedspace } = setupBookingStateStubs(booking)
-
-    const bookingShow = BookingShowPage.visit(premises, bedspace, booking)
-    bookingShow.clickExtendBookingButton()
-
-    const extensionPage = Page.verifyOnPage(BookingExtensionNewPage, premises, bedspace, booking)
-
-    const tooLateDepartureDate = DateFormats.dateObjToIsoDate(addDays(booking.departureDate, 85))
-
-    const extension = newExtensionFactory.build({ newDepartureDate: tooLateDepartureDate })
-    extensionPage.completeForm(extension)
-
-    const newOverstay = Cas3NewOverstay.build({ newDepartureDate: tooLateDepartureDate })
-    Page.verifyOnPage(BookingOverstayNewPage, premises, bedspace, booking, newOverstay)
   })
 
   it('allows me to record an overstay', () => {
@@ -56,7 +39,7 @@ context('Booking overstay', () => {
     extensionPage.completeForm(extension)
 
     const overstay = cas3OverstayFactory.build({ newDepartureDate: tooLateDepartureDate })
-    const overstayPage = Page.verifyOnPage(BookingOverstayNewPage, premises, bedspace, booking, overstay)
+    const overstayPage = Page.verifyOnPage(BookingOverstayNewPage, premises, bedspace, booking, overstay, false)
 
     const newOverstay = cas3NewOverstayFactory.build({ ...overstay })
     cy.task('stubOverstayCreate', { premisesId: premises.id, bookingId: booking.id, overstay: newOverstay })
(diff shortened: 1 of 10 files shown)


integration_tests/tests/temporary-accommodation/manage/overstay.cy.ts
```diff
@@ -37,7 +40,29 @@ context('Booking overstay', () => {
     extensionPage.completeForm(extension)
 
     const newOverstay = Cas3NewOverstay.build({ newDepartureDate: tooLateDepartureDate })
-    Page.verifyOnPage(BookingOverstayNewPage, premises, bedspace, booking, newOverstay)
+    Page.verifyOnPage(BookingOverstayNewPage, premises, bedspace, booking, newOverstay, false)
+  })
+
+  it('navigates to the overstay via the departure page', () => {
```
froddd: This test and the one above (`'navigates to the overstay page via the booking extension page'`) aren't needed, as all those same steps are covered in the subsequent related tests, `'allows me to record an overstay'` and `'allows me to record an overstay for a departure'`. This can save a bit of time running integration tests!

server/controllers/temporary-accommodation/manage/departuresController.test.ts
```diff
@@ -127,6 +133,44 @@ describe('DeparturesController', () => {
...
+      await requestHandler(request, response, next)
+
+      const address = urlFormat({
+        pathname: paths.bookings.overstays.new({ premisesId, bedspaceId, bookingId }),
+        query: { newDepartureDate },
+      })
+
+      expect(bookingService.getBooking).toHaveBeenCalledWith(callConfig, premisesId, bookingId)
+      expect(response.redirect).toHaveBeenCalledWith(address)
+      expect(request.session.departure).toEqual(expect.objectContaining(newDeparture))
```
froddd, line 171: Should this be full equality, i.e.:
```suggestion
      expect(request.session.departure).toEqual(newDeparture)
```

server/controllers/temporary-accommodation/manage/overstaysController.test.ts
```diff
@@ -91,6 +97,46 @@ describe('OverstaysController', () => {
...
+      bedspaceService.getSingleBedspace.mockResolvedValue(bedspace)
+      bookingService.getBooking.mockResolvedValue(booking)
+
+      const requestHandler = overstaysController.new()
+      ;(fetchErrorsAndUserInput as jest.Mock).mockReturnValue({ errors: {}, errorSummary: [], userInput: {} })
+
+      await requestHandler(request, response, next)
+
+      expect(premisesService.getSinglePremises).toHaveBeenCalledWith(callConfig, premises.id)
+      expect(bedspaceService.getSingleBedspace)
```
froddd: This is missing an assertion, I assume `.toHaveBeenCalledWith(callConfig, ...)`?

server/controllers/temporary-accommodation/manage/overstaysController.ts
```diff
@@ -85,18 +100,34 @@ export default class OverstaysController {
...
+            redirectUrl = paths.bookings.departures.new({ premisesId, bedspaceId, bookingId })
+            throw error
+          }
+        }
+
         await this.overstaysService.createOverstay(callConfig, premisesId, bookingId, newOverstay)
 
-        req.flash('success', 'Booking departure date changed')
+        if (departure !== undefined) {
+          redirectUrl = paths.bookings.departures.new({ premisesId, bedspaceId, bookingId })
```
froddd, line 113: This updated `redirectUrl` isn't actually used anywhere, and the integration test seems to indicate the user should be redirected to the Booking page anyway -- if so, this can go, or the integration test and implementation should be updated!

server/controllers/temporary-accommodation/manage/overstaysController.ts
```diff
@@ -85,18 +100,34 @@ export default class OverstaysController {
...
+      let successMessage = 'Booking departure date changed'
+
       try {
+        const { departure } = req.session
+        if (departure !== undefined) {
+          if (DateFormats.isoToDateObj(departure.dateTime) > new Date()) {
+            const error = new Error()
+            insertGenericError(error, 'dateTime', 'departureDateInFuture')
+            redirectUrl = paths.bookings.departures.new({ premisesId, bedspaceId, bookingId })
+            throw error
```
froddd: Should this not be validated in the `departuresController` instead? It seems like there's little point letting the user complete the overstay information if they've entered a date of departure that is not going to be accepted.

server/controllers/temporary-accommodation/manage/overstaysController.test.ts
```diff
@@ -195,6 +273,69 @@ describe('OverstaysController', () => {
...
+
+      expect(catchValidationErrorOrPropogate).toHaveBeenCalledWith(
+        request,
+        response,
+        err,
+        paths.bookings.departures.new({ premisesId: premises.id, bedspaceId: bedspace.id, bookingId: booking.id }),
+      )
+    })
+
+    it('redirects to the departure page with errors if the departure is in the future', async () => {
```
froddd: I think this should be handled in the `departuresController`, as per comment in `overstaysController.ts`. However, this controller should ideally also handle API errors, so if the API returns an error with departure date being in the future, this handler should redirect the user to the correct form page...

server/controllers/temporary-accommodation/manage/overstaysController.ts
```diff
@@ -85,18 +100,34 @@ export default class OverstaysController {
...
+          }
+        }
+
         await this.overstaysService.createOverstay(callConfig, premisesId, bookingId, newOverstay)
```
froddd, line 109: It might be an edge case, but what would happen if the `createOverstay()` call was successful but the `createDeparture()` one below wasn't? I'm not sure there's a solution to this without re-engineering the API to accept a departure as part of the creation of an overstay (or vice-versa)...

froddd (commented):
Nice work!

There's some duplication in integration tests, which I think appeared in the first PR for overstays -- I should have caught that! It means the tests can be sped up a little though, which is always good.

A bit of a rejig might be needed to handling of date validation for departures: at the moment entering a future date that results in an overstay only tells the user the date cannot be in the future _after_ they've completed the overstay form, which isn't ideal. It would help if the UI validation was moved to the departures controller to prevent progressing to the next form; however, the overstays controller should also handle that error being returned by the API, and redirect the user correctly in that edge case!

server/controllers/temporary-accommodation/manage/overstaysController.ts
robanderson-moj: The redirectUrl is used in the `catch` block, if the backend call to create the departure fails we need to redirect back to the new departure page
It isn't a wonderful solution, but it's the best I could come up with without needing a new backend endpoint or duplicated views

server/controllers/temporary-accommodation/manage/departuresController.test.ts
robanderson-moj: I think this was because the departure stored in the session was created using the a spread, so contains individual date component fields as well. I'm checking here that the expected `newDeparture` is a subset of the object containing the date fields

froddd: I see: the controller sets the `request.session.departure` value to an object of type `Cas3NewDeparture`, so it _should_ be full equality, but, the controller also spreads the whole of the `request.body` and the full output of `DateFormats.dateAndTimeInputsToIsoString` into this object, which means it then also carries the date parts, which is what fails the test if using strict equality. Let's leave it as it is. 

I'm not sure why typecheck isn't picking up on the type discrepancies though! One can even add whatever other properties to the assignment as `Cas3NewDeparture` and nothing complains. Odd.

server/controllers/temporary-accommodation/manage/overstaysController.ts
froddd: I see -- I was wondering why it showed as unused in the IDE. If it works!

server/controllers/temporary-accommodation/manage/overstaysController.ts
robanderson-moj: Interestingly this validation was being done after the call to `departureService.createDeparture` in the `departuresController`, so I've just changed the order around a little to reduce the duplication.

server/controllers/temporary-accommodation/manage/overstaysController.ts
robanderson-moj: I did test this by purposely returning an error from the BE, it's possible to create a new overstay over the top of the old one, so if there's a random error in `createDeparture` it can be repeated without any weird side effects

server/controllers/temporary-accommodation/manage/departuresController.test.ts
froddd: TIL (well, yesterday): if one adds a spread object cast as `any` to a typed object, it's considered valid whatever members are added:

```typescript
type X = { foo: string }

// This is obviously not valid because type X does not have member `bar`:
const test1: X = {
  ...({ foo: 'qux' }),
  bar: 0
}

// But this is valid 🤯 
const test2: X = {
  ...({ foo: 'qux' } as any),
  bar: 0
}
```

Which explains why the typechecking does not complain about spreading date parts into an object of type `Cas3NewDeparture` -- type of `req.body` is `any` 🤷‍♂️

server/controllers/temporary-accommodation/manage/departuresController.ts
```diff
@@ -55,14 +56,46 @@ export default class DeparturesController {
...
+            insertGenericError(error, 'moveOnCategoryId', 'empty')
+          }
+          throw error
+        }
 
         if (DateFormats.isoToDateObj(newDeparture.dateTime) > new Date()) {
           const error = new Error()
           insertGenericError(error, 'dateTime', 'departureDateInFuture')
           throw error
         }
```
froddd, line 77: If a future departure date is submitted but no reason is selected, this will only show an error for the reason ID, then show the error for the future date once the reason ID is selected and resubmitted.

All the errors can be checked at once based on whether `error` has a `data` property (added by any call to `insertGenericError` -- useful side effect 😬):


```suggestion
// import { SanitisedError } from '../../../sanitisedError'

       const error = new Error()

        if (!newDeparture.dateTime) {
          insertGenericError(error, 'dateTime', 'empty')
        } else if (DateFormats.isoToDateObj(newDeparture.dateTime) > new Date()) {
          insertGenericError(error, 'dateTime', 'departureDateInFuture')
        }
        if (!newDeparture.reasonId) {
          insertGenericError(error, 'reasonId', 'empty')
        }
        if (!newDeparture.moveOnCategoryId) {
          insertGenericError(error, 'moveOnCategoryId', 'empty')
        }

        if ((error as SanitisedError).data) {
          throw error
        }
```

froddd (approved):
Nice one!

A small suggestion for handling all potential new departure errors at once ,but good to go!

PR #1556, opened 2026-02-02
Chore: improve maintenance mode

froddd:
# Changes in this PR

This PR updates the mainetnance mode:
- Removes ambiguity for config value, now always returned as a `Boolean`.
- Ensures users accessing the `/maintenance` url get redirected to the root URL `/` if maintenance mode is disabled -- this will ensure users refreshing the maintenance page after the maintenance mode has been disabled can continue using the service.
- Adds missing tests to cover maintenance page redirect implementation.


robanderson-moj (approved):
👍

PR #1564, opened 2026-02-06
Added support for overstay when editing departures

robanderson-moj:
# Context

Still related to [CAS-2140](https://dsdmoj.atlassian.net/browse/CAS-2140), adding the overstay flow when editing/updating a departure over 84 nights.

Thankfully updating a departure just POSTs a new departure over the top of the old one, so I was able to re-use a lot of the work from adding overstays to new departures with a minor refactor.

I've also refactored some of the error handling in the `overstaysController` based on some of Fred's feedback from my last PR

### DIFF
--- a/server/controllers/temporary-accommodation/manage/departuresController.ts
+++ b/server/controllers/temporary-accommodation/manage/departuresController.ts
@@ -1,13 +1,19 @@
 import type { Request, RequestHandler, Response } from 'express'
-import type { Cas3NewDeparture } from '@approved-premises/api'
 import { format as urlFormat } from 'url'
 import paths from '../../../paths/temporary-accommodation/manage'
 import { BookingService, DepartureService, PremisesService } from '../../../services'
 import BedspaceService from '../../../services/bedspaceService'
-import { DateFormats, nightsBetween } from '../../../utils/dateUtils'
+import {
+  dateAndTimeInputsAreValidDates,
+  DateFormats,
+  dateIsBlank,
+  dateIsInFuture,
+  nightsBetween,
+} from '../../../utils/dateUtils'
 import extractCallConfig from '../../../utils/restUtils'
 import { catchValidationErrorOrPropogate, fetchErrorsAndUserInput, insertGenericError } from '../../../utils/validation'
 import config from '../../../config'
+import { SanitisedError } from '../../../sanitisedError'
 
 export default class DeparturesController {
   constructor(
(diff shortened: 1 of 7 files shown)


server/controllers/temporary-accommodation/manage/departuresController.ts
```diff
@@ -170,4 +175,36 @@ export default class DeparturesController {
       }
     }
   }
+
+  private validateNewDeparture(newDeparture: Cas3NewDeparture): void {
+    const error = new Error()
+
+    if (!newDeparture.dateTime) {
+      insertGenericError(error, 'dateTime', 'empty')
```
bobmeredith, line 188: `insertGenericError` is using a side-effect which is jolly bad form IMHO.
It makes how `error` gets mutated a bit obscure.

bobmeredith (dismissed):
Looks good.
Just a tiny comment - feel free to ignore.

server/controllers/temporary-accommodation/manage/departuresController.ts
```diff
@@ -170,4 +175,36 @@ export default class DeparturesController {
...
+    }
+    if (!newDeparture.reasonId) {
+      insertGenericError(error, 'reasonId', 'empty')
+    }
+    if (!newDeparture.moveOnCategoryId) {
+      insertGenericError(error, 'moveOnCategoryId', 'empty')
+    }
+
+    if ((error as SanitisedError).data) {
+      throw error
```
froddd, line 204: Throwing here means the next 2 validators may not run until the first errors are fixed -- this isn't great for UX. All validators should run on the same request, so the user doesn't have to submit the form multiple times.

robanderson-moj: I'm under the impression that this is how the backend would have handled validation previously, but I'll add an `else` block after the `dateTime` check to handle cases where the the `dateTime` is present but incorrectly formatted.

After refactoring, it's a little less easy to follow, but it's possible to get a missing reason or moveOnCategory error while also getting an error that the date is incorrectly formatted or is in the future

froddd: There is a validator for date input fields (`dateAndTimeInputsAreValidDates`) which would avoid the `try/catch` inside the conditional, but it should validate the request body, before assigning anything to `newDeparture`. Which sounds a tad more logical, in any case: validate, then, if there is no error, prepare data to submit. This looks ok though -- as long as it validates at once and doesn't impact user experience, this is better.

robanderson-moj: I've ended up using `dateAndTimeInputsAreValidDates` to validate, and the `newDeparture` is only instantiated once the validation has been performed on the body of the request. The only thing I'm not happy about is passing the request to the validation method rather than the body, but this was primarily to avoid passing an object with type `any`.

froddd (approved):
Looks good!

PR #1601, opened 2026-03-26
Adding SQS variables to test helm chart to fix test deployment

robanderson-moj:
# Context

Part of ticket https://dsdmoj.atlassian.net/browse/FM-426

I've already deployed this branch to the `test` environment, so it has fixed the issue we were seeing previously

The issue was a consequence of the audit logging we updated recently, as the environment variable `get` function in the `hmpps-audit-client` had different logic that was throwing an exception if `NODE_ENV` was set to `production` and the environment variables for the SQS variables weren't set


bobmeredith (approved):
LGTM

PR #1611, opened 2026-04-01
Add scripts to split Cypress E2E and integration tests

robanderson-moj:
# Context

Re-add test splitting for E2E tests using a bash script rather than `chaosaffe/split-tests`

Also split the integration tests in 2 in order to speed up the build process by around 4 minutes

# Changes in this PR

### Before
image

### After
image


AvraamK (approved):
🚀

PR #1614, opened 2026-04-15
Chore: Fix axios and diff vulnerabilities

froddd:
This overrides the versions of `axios` and `diff` to ensure no vulnerable version is used.


robanderson-moj (approved):
👍

PR #1638, opened 2026-05-11
Chore/use cas1 shared actions

froddd:
This PR replaces local scripts and actions with the shared actions from the CAS1 repository, both for [splitting tests](https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/main/.github/actions/split_tests/action.yml) and sending a [Slack Failure Notification](https://github.com/ministryofjustice/hmpps-approved-premises-ui/blob/main/.github/actions/slack_failure_notification/action.yml). The actions are versioned on the CAS1 repository, so any update will be picked up by Renovate when a new tag is created on the CAS1 repo.


robanderson-moj (approved):
👍 
Nice removing both of the split scripts here too

PR #1656, opened 2026-06-09
FM-711 Remove unused ‘applicable tier’ logic

davidatkinsuk:
This appears to have been copied from CAS1, but isn’t used (there is nothing calling the `isApplicableTier` function in `personUtils.ts`


robanderson-moj (approved):
Can't find anywhere in the codebase where `isApplicableTier` is used, and all of the E2E tests still pass - happy to approve

